PR #9, opened 2022-04-18
Added a "help" player command and "sendmsg" server command. Separated commands and aliases 

4Benj:
Changes:
- Help command for the player
     help me pls (Might be an issue later with the auto-scroll within genshin's chat)
- sendmsg command
     Sends a message to the player
     sendmsg [player id] [message]
- Players can now send a command using '/' and '!' 
     I did this because the helpText for existing commands seemed to assume that the commands used '/'.
- getAccountByPlayerId(int playerID) is now public within DatabaseHelper.
     This was for sendmsg. I don't know why getAccountById would be public and getAccountByPlayerId isn't. If there was a better way to do this, I'll happily change sendmsg to use that and revert my change.
- Separated commands and aliases into different HashMaps. It seems more logical not to have them grouped together (It also declutters the help command). I do have another way of doing this with one HashMap but I think it's better like this

### DIFF
--- a/src/main/java/emu/grasscutter/commands/PlayerCommands.java
+++ b/src/main/java/emu/grasscutter/commands/PlayerCommands.java
@@ -4,6 +4,7 @@ import java.lang.reflect.Modifier;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


lunaticwhat:
seg

PR #20, opened 2022-04-19
Give character command

SpikeHD:
A command for adding characters to a players roster.

### DIFF
--- a/src/main/java/emu/grasscutter/commands/PlayerCommands.java
+++ b/src/main/java/emu/grasscutter/commands/PlayerCommands.java
@@ -176,6 +176,35 @@ public final class PlayerCommands {
             }
         }
     }
+
+    @Command(label = "givechar", aliases = { "givec" }, usage = "Usage: givechar [avatar id] [level]")
+    public static class GiveCharCommand implements CommandHandler {
+
+        @Override
+        public void execute(GenshinPlayer player, List<String> args) {
+            int avatarID = Integer.parseInt(args.get(0));
+            int level = Integer.parseInt(args.get(1));
+
+            if(player == null) {
+                CommandHandler.sendMessage(null, "Player not found."); return;
+            }
+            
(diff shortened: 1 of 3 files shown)


SpikeHD:
Not done yet gimme a sec to bring it up-to-speed with the conventions for everything else

SpikeHD:
Alright, tested and ready :)

src/main/java/emu/grasscutter/commands/PlayerCommands.java
```diff
@@ -176,6 +177,110 @@ public void execute(GenshinPlayer player, List<String> args) {
...
         }
     }
+
+    @Command(label = "givechar", aliases = { "givec" }, usage = "Usage: givechar <playerId> <avatarId> [level]")
+    public static class GiveCharCommand implements CommandHandler {
+        @Override public void execute(GenshinPlayer player, List<String> args) {
+            int target, avatarID, level = 1;
+
+            if(args.size() < 2) {
+                CommandHandler.sendMessage(null, "Usage: give <player> <avatarId> [level]");
```
MlgmXyysd, line 187: `give` instead of `givechar`?

4Benj: Maybe make the player variable optional when the player is submitting the command

SpikeHD: > `give` instead of `givechar`?

Good catch, that's fixed!

MlgmXyysd: > Maybe make the player variable optional when the player is submitting the command

makes sense

Melledy:
When directly setting levels on avatars, you need to also take in account their ascension level and ascension talents.

SpikeHD:
> ascension talents

Good call, I will make sure that's added

SpikeHD:
Both ascension levels and talents should be properly given and calculated now

PR #22, opened 2022-04-19
Set world level command

SpikeHD:
Simple command to set world level. Does so in both the player properties *and* world properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ There is a dummy user named "Server" in every player's friends list that you can
 
 `!killall`
 
+`!setworldlevel [level]` - Relog to see effects properly
+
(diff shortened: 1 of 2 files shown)


SpikeHD:
Two things:

1. Is it necessary to limit to the actual max world level of the game? Does seem to hurt anything going past
2. After setting to something above 5, the option to lower world level does not appear. I am not totally sure why this is so if someone has insight lemme know

memetrollsXD:
For now, it isn't.. but if natural enemy spawning gets added in the future, prop values over level 200 don't exist anymore so you'd run into some issues, if you can cap the mob level, that's fine then

PR #24, opened 2022-04-19
Fix weapon refine

Yazawazi:
Sometimes, for example, a weapon with ID `11415` has a refined material that is not itself but is based on its `AwakenMaterial` value (Some will be available, in `WeaponExcelConfigData.json`).


Yazawazi:
Oh, sorry, I forgot to change this:

`player.getInventory().removeItem(feed);` -> `player.getInventory().removeItem(feed, 1);` in `refineWeapon` method. （＞人＜；）

PR #31, opened 2022-04-19
Added bonus for ingame function

alt3ri:
A tip to teleport since TP waypoint didn't work.


lunaticwhat:
useful info so. Should be in the wiki later @KingRainbow44

PR #33, opened 2022-04-19
Fix folder separation for folderNames

lilmayofuksu:


lunaticwhat:
so true

Melledy:
Don’t need to add a Utils::toFilePath call in your folder array since it just gets called anyways when iterating.

lilmayofuksu:
> Don’t need to add a Utils::toFilePath call in your folder array since it just gets called anyways when iterating.

somebody forgot to edit those \\ then

lilmayofuksu:
bruh

PR #36, opened 2022-04-19
Automatically detect java and mitmproxy

w4123:
Just to make running the server a bit easier

### DIFF
--- a/run.cmd
+++ b/run.cmd
@@ -10,11 +10,40 @@ call :LOG [INFO] Initializing...
 
 @rem This will not work if your java or mitmproxy is in a different location, plugin as necessary
 @rem this just saves you from changing your PATH
-set JAVA_PATH=C:\Program Files\Java\jdk1.8.0_202\
-set MITMPROXY_PATH=%~dp0
+set JAVA_PATH=
+set MITMPROXY_PATH=
 set PROXY_SCRIPT=proxy
 @rem TODO: MongoDB integration
 set SERVER_PATH=%~dp0
+
+@rem detect java
+setlocal enabledelayedexpansion
+if "%JAVA_PATH%" == "" (
+	if not "%JAVA_HOME%" == "" (
+		set JAVA_PATH=%JAVA_HOME%
+	)
+)
+
+if "%JAVA_PATH%" == "" (
+	where java >nul 2>nul
+	if "!ERRORLEVEL!" == "0" (
+		for /f "usebackq tokens=*" %%F in (`where java`) do ( 
+			set JAVA_PATH=%%F
+		)
+		set JAVA_PATH=!JAVA_PATH:~0,-12!
+	)
+)
+
+@rem detect mitmproxy
+if "%MITMPROXY_PATH%" == "" (
+	where mitmdump >nul 2>nul
+	if "!ERRORLEVEL!" == "0" (
+		for /F "usebackq tokens=*" %%F in (`where mitmdump`) do ( 
+			set MITMPROXY_PATH=%%F
+		)
+		set MITMPROXY_PATH=!MITMPROXY_PATH:~0,-12!
+	)
+)
 
 @rem mitmproxy not found, server only
 if not exist "%MITMPROXY_PATH%mitmdump.exe" (
@@ -107,6 +136,7 @@ if "%PROXY%" == "" (
 
 	certutil -delstore root %SERIAL% >nul 2>nul
 )
+setlocal disabledelayedexpansion
 
 call :LOG [INFO] See you again :)
 goto :EOF


run.cmd
```diff
@@ -10,12 +10,41 @@ call :LOG [INFO] Initializing...
...
+	if not "%JAVA_HOME%" == "" (
+		set JAVA_PATH=%JAVA_HOME%
+	)
+)
+
+if "%JAVA_PATH%" == "" (
+	where java >nul 2>nul
+	if "!ERRORLEVEL!" == "0" (
+		for /f "usebackq tokens=*" %%F in (`where java`) do ( 
+			set JAVA_PATH=%%F
```
MlgmXyysd, line 31: No, we need a specific Java version: `1.8.0u202b08`

MlgmXyysd (changes requested):

run.cmd
w4123: I understand but this does not harm anyway...
Maybe add a version check at the same time?

MlgmXyysd: > I understand but this does not harm anyway... Maybe add a version check at the same time?

I think `where` will only search for files in `PATH`, so it's useless
image
image

w4123: It will also search in the current directory if I remember correctly

MlgmXyysd: > It will also search in the current directory if I remember correctly

But no, my `mitmdump` is in `D:\Grasscutter`, it didn't help me find it. There are 3 `java.exe` under `C:\Program Files\Java`, it didn't find too.

w4123: Maybe let's keep the mitmproxy part first

MlgmXyysd: Maybe it's a good idea to get the installation path from the registry. But if the registry has no installation information, the automatic search still does not work.

w4123: > > It will also search in the current directory if I remember correctly
> 
> But no, my `mitmdump` is in `D:\Grasscutter`, it didn't help me find it. There are 3 `java.exe` under `C:\Program Files\Java`, it didn't find too.

non-recursively...

MlgmXyysd: > Maybe let's keep the mitmproxy part first

As you see, it doesn't work either.🤔

w4123: > Maybe it's a good idea to get the installation path from the registry. But if the registry has no installation information, the automatic search still does not work.

lemme check

w4123: > > Maybe let's keep the mitmproxy part first
> 
> As you see, it doesn't work either.🤔

well, at least it works when mitmdump is in PATH

MlgmXyysd: > Maybe it's a good idea to get the installation path from the registry. But if the registry has no installation information, the automatic search still does not work.

Since `mongod` and `mitmdump` (even `java`) can be run directly without installation, detecting them is still a big problem if users use ZIP methods.
(https://github.com/Melledy/Grasscutter/pull/30#issuecomment-1102702443)

The easiest way is not to detect, but to put these directly with Grasscutter.

MlgmXyysd: > > > Maybe let's keep the mitmproxy part first
> > 
> > 
> > As you see, it doesn't work either.🤔
> 
> well, at least it works when mitmdump is in PATH

If it's in `PATH`, you don't need to fill the full path

w4123: > > > > Maybe let's keep the mitmproxy part first
> > > 
> > > 
> > > As you see, it doesn't work either.🤔
> > 
> > 
> > well, at least it works when mitmdump is in PATH
> 
> If it's in `PATH`, you don't need to fill the full path

The current script doesn't really work when mitmproxy is in PATH ... because it assume it is located at current directory

w4123: The location can be retrieved from
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{26A24AE4-039D-4CA4-87B4-2F64180202F0}\InstallLocation
but yes, it won't work if it is not installed by exe

MlgmXyysd: > > > > > Maybe let's keep the mitmproxy part first
> > > > 
> > > > 
> > > > As you see, it doesn't work either.🤔
> > > 
> > > 
> > > well, at least it works when mitmdump is in PATH
> > 
> > 
> > If it's in `PATH`, you don't need to fill the full path
> 
> The current script doesn't really work when mitmproxy is in PATH ... because it assume it is located at current directory

Yes, so you can keep it blank when miymproxy in PATH: `set MITMPROXY_PATH=` These variables are designed to be modifiable.

w4123: > The location can be retrieved from HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall{26A24AE4-039D-4CA4-87B4-2F64180202F0}\InstallLocation but yes, it won't work if it is not installed by exe

Shall we try this or shall we just insist that the user should modify the script

MlgmXyysd: > > The location can be retrieved from HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall{26A24AE4-039D-4CA4-87B4-2F64180202F0}\InstallLocation but yes, it won't work if it is not installed by exe
> 
> Shall we try this or shall we just insist that the user should modify the script

I think the existing should be kept. I'll think about it during the daytime again, now time to sleep.

w4123: > > > The location can be retrieved from HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall{26A24AE4-039D-4CA4-87B4-2F64180202F0}\InstallLocation but yes, it won't work if it is not installed by exe
> > 
> > 
> > Shall we try this or shall we just insist that the user should modify the script
> 
> I think the existing should be kept. I'll think about it during the daytime again, now time to sleep.

Good night and thanks for your suggestion. I will close the PR for now.

MlgmXyysd: > > > > The location can be retrieved from HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall{26A24AE4-039D-4CA4-87B4-2F64180202F0}\InstallLocation but yes, it won't work if it is not installed by exe
> > > 
> > > 
> > > Shall we try this or shall we just insist that the user should modify the script
> > 
> > 
> > I think the existing should be kept. I'll think about it during the daytime again, now time to sleep.
> 
> Good night and thanks for your suggestion. I will close the PR for now.

Good night :)

w4123: > > > > > > Maybe let's keep the mitmproxy part first
> > > > > 
> > > > > 
> > > > > As you see, it doesn't work either.🤔
> > > > 
> > > > 
> > > > well, at least it works when mitmdump is in PATH
> > > 
> > > 
> > > If it's in `PATH`, you don't need to fill the full path
> > 
> > 
> > The current script doesn't really work when mitmproxy is in PATH ... because it assume it is located at current directory
> 
> Yes, so you can keep it blank when mitmproxy in PATH: `set MITMPROXY_PATH=` These variables are designed to be modifiable.

FYI, setting the variable to empty does not work even if the executable is in PATH as "if not exists" statement does not check in PATH.

PR #37, opened 2022-04-19
Restart command

memetrollsXD:
Implements a command to easily "restart" the session, so the login sequence happens over again. Useful if you edited the MongoDB database manually and don't want to restart

For now uses getSession().close(), which will make the client linger for a few seconds before re-connecting


memetrollsXD:
@KingRainbow44 Ready for merging!

PR #39, opened 2022-04-20
Wiki: How to edit character level and etc.

ghost:
Don't know how to PR to Wiki, so using _actuallyeunha_'s method


KingRainbow44:
Instead of merging it as a file, I'll be adding this to the wiki. (user request)

PR #40, opened 2022-04-20
Create the account if it doesn't exist on the login screen

KingRainbow44:
This is more for *public* instances of Grasscutter, where every user can't be created via. the server console.

### DIFF
--- a/src/main/java/emu/grasscutter/commands/CommandMap.java
+++ b/src/main/java/emu/grasscutter/commands/CommandMap.java
@@ -113,7 +113,7 @@ public final class CommandMap {
         if(player != null) {
             String permissionNode = this.annotations.get(label).permission();
             Account account = player.getAccount();
-            if(!permissionNode.equals("*") && !account.hasPermission(permissionNode)) {
+            if(!permissionNode.isEmpty() && !account.hasPermission(permissionNode)) {
                 CommandHandler.sendMessage(player, "You do not have permission to run this command."); return;
             }
         }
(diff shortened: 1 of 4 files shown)


src/main/java/emu/grasscutter/commands/CommandMap.java
```diff
@@ -113,7 +113,7 @@ public void invoke(GenshinPlayer player, String rawMessage) {
         if(player != null) {
             String permissionNode = this.annotations.get(label).permission();
             Account account = player.getAccount();
-            if(permissionNode != "" && !account.hasPermission(permissionNode)) {
+            if(!Objects.equals(permissionNode, "") && !account.hasPermission(permissionNode)) {
```
MlgmXyysd: `permissionNode.equals("")`

MlgmXyysd (changes requested):

src/main/java/emu/grasscutter/commands/CommandMap.java
```diff
@@ -113,7 +113,7 @@ public void invoke(GenshinPlayer player, String rawMessage) {
         if(player != null) {
             String permissionNode = this.annotations.get(label).permission();
             Account account = player.getAccount();
-            if(permissionNode != "" && !account.hasPermission(permissionNode)) {
+            if(!permissionNode.equals("*") && !account.hasPermission(permissionNode)) {
```
MlgmXyysd: `"*"`?🧐

KingRainbow44: did i mess up something lmao

ghost (approved):

PR #41, opened 2022-04-20
Try to implement NpcTalk packets

w4123:
Currently, the client will pause for a long time after talking to an NPC. This is because the NpcTalk* packets are not implemented.
This PR tries to implement these packets, only a stub packet is included for now but at least it removes the long pause.

Work in Progress to ensure correctness and completeness.


w4123:
Didn't find any information about which item to give after a conversation with rewards... does anyone has any idea?

memetrollsXD:
You need to do that server-side with the NPC id, and then add that to a list saved in the database, so you can send MetNpcListNotify at startup

lunaticwhat:
a weird information: If during the hang after talking to NPCs, by reconnecting to the server made the client can continue after that hang with dialogues or anything that a normally the game would work.

w4123:
> You need to do that server-side with the NPC id, and then add that to a list saved in the database, so you can send MetNpcListNotify at startup

I tried to play with the MetNpcListNotify packet but it seems that it is not really related to talking to Npcs.  MetNpcListNotify, MeetNpcReq and MeetNpcRsp are used to determine and record whether an Npc has been met before. I tried to set it on random Npcs but it yield no effect. According to NpcFirstMetExcelConfigData.json, I suspect that the field is only used for a very small proportion of Npcs but not all.

Therefore, the problem of giving items to the player after they complete the dialogue with rewards remains unsolved. I think maybe a stub that prevents the hang of the client when talking to the Npcs would be sufficient for the moment. I will investigate deeper and try to implement the complete feature in another PR or something.

memetrollsXD:
MeetNpc is used for NPCs that give rewards, so it doesn't give them twice

w4123:
> 

There is no MeetNpcReq packet sent from the client when I complete the dialogue with an NPC that gives rewards...
Meanwhile, I don't know which item to give anyway.

ghost:
@w4123 Could you make a pull with the protos you added into [https://github.com/Grasscutters/Grasscutter-Protos](https://github.com/Grasscutters/Grasscutter-Protos)? Thanks!

w4123:
> @w4123 Could you make a pull with the protos you added into https://github.com/Grasscutters/Grasscutter-Protos? Thanks!

https://github.com/Grasscutters/Grasscutter-Protos/pull/1

PR #42, opened 2022-04-20
Seperate commands list

SpikeHD:
As the command list grows, I think it will become a bit too heavy to include in the main README, so i thought it would be a good idea to move it instead.

I also put them into a table which I thought looked a bit nicer.

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -1,0 +1,27 @@
+# Table of Contents
+
+* [Server console commands](#server-console-commands)
+* [Client/In-game commands](#in-game-commands)
+
+
+### Server console commands
+
+|Command|Function|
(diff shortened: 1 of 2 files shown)


MlgmXyysd:
Have you all forgotten [wiki](https://github.com/Grasscutters/Grasscutter/wiki)

SpikeHD:
I actually don't know how to send a PR for the wiki so uhhh if anyone wants to do that feel free

lunaticwhat:
yeah that was what I ve talk about ! we should have a command list in wiki and link to it on the readme

alt3ri:
true

ghost:
Please make the PR to the development branch

PR #55, opened 2022-04-20
Custom Y position on MarkMapReq

memetrollsXD:
If no Y provided in MapMark name, default to 300 (old value)


memetrollsXD:
This is pretty stable too, but I don't know the requirements to merge to stable, so I just did development instead

PR #74, opened 2022-04-21
Server run modes and multi-server dispatch support

4Benj:
Changes: 
- The server can now be ran in HYBRID (original), DISPATCH_ONLY, or GAME_ONLY modes, which as the name suggests turns on and off the dispatch and game servers depending on its value in the config
- Multiple game servers can be configured in the server's configured so players can connect to different servers setup by server administrators.
- Completely overhauled the config to move things into dispatch or game server only sections

Things that need to be improved:
- Currently the game servers does not connect to the dispatch server to communicate. In order to authenticate the user, game servers need access to the dispatch database. The current plan is for creating an authentication route in the dispatch server to checking the user's session token. 

image


4Benj:
Have just realised there is an issue in HYBRID mode where the user data isn't stored correctly on the dispatch server.
Fix is currently being attempted.

4Benj:
I had to get rid of my users hashmap in the dispatch server as it wasn't doing what I intended at all.
The only solution I can see now is connecting the game server to the dispatch server which will have to be done anyway 😢

4Benj:
Other then that though everything works and is safe to merge. Unless the maintainers want the servers to be connected first.

Mar0xy:
IMO The servers should be connected first for simplicity reasons instead of complicating it further down the line but hey that's just my opinion.

4Benj:
All the issues with `PlayerLoginRsp` should now be fixed 👍
It's ready to be merged but I'll probably do another PR another day adding routes to the dispatch server that verifies the player's session, so then we don't have to have two databases on one server.

ghost (approved):

PR #108, opened 2022-04-22
GM handbook: Explicitly load and save UTF-8

fumbling644o:
This fixes potential output issues for non-Latin scripts when exporting the handbook by changing the filename in the code. This is likely only a problem for Windows users.


ghost (approved):

iTruth:
for java8, there's no such a constructor  
https://docs.oracle.com/javase/8/docs/api/java/io/FileReader.html

PR #119, opened 2022-04-22
Revert "GM handbook: Explicitly load and save UTF-8"

Scirese:
Reverts Grasscutters/Grasscutter#108

for java8, there's no such a constructor
https://docs.oracle.com/javase/8/docs/api/java/io/FileReader.html


ghost:
Already reverted.

PR #125, opened 2022-04-22
fixed encoding problem when generating gm handbook

PasserbyAlpha:
I've met the same problem as the message on [discord](https://discord.com/channels/965284035985305680/965284035985305688/965832388716404787), which is the result of default encoding in java.

When using FileReader, the default encoding will be applied. However, in some area utf8 is not the default charset, which will result in exception when loading text map file.

To solve the problem, InputStreamReader is applied in this branch, and charset is manually set to utf8. I've tested that works on my computer.

### DIFF
--- a/src/main/java/emu/grasscutter/tools/Tools.java
+++ b/src/main/java/emu/grasscutter/tools/Tools.java
@@ -1,9 +1,11 @@
 package emu.grasscutter.tools;
 
 import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


fumbling644o:
Hello! Please consider changing line 38 to this, so the output is also UTF-8:

`try (PrintWriter writer = new PrintWriter(new OutputStreamWriter(new FileOutputStream(fileName), StandardCharsets.UTF_8), false)) {`

My IDE is set to the correct language version this time :)

PasserbyAlpha:
> Hello! Please consider changing line 38 to this, so the output is also UTF-8:
> 
> `try (PrintWriter writer = new PrintWriter(new OutputStreamWriter(new FileOutputStream(fileName), StandardCharsets.UTF_8), false)) {`
> 
> My IDE is set to the correct language version this time :)



> fumbling644o

That's great! It works great on non-EN text maps (*^▽^*)

I've applied this to the latest commit

ghost (approved):

PR #126, opened 2022-04-22
Add PublicPort field to Dispatch server

omg-xtao:
When using `nginx or other reverse proxy` tool, the `public port` may not match the original port. This will cause the client to report error 4206.

So add a `PublicPort` field to make server work when two ports do not match.


MlgmXyysd:
Port forwarding can also cause this problem.

BTW, I think the Dispatch server's port might have been redirected by the proxy daemon. The main issue might be the game server, so it's best to add it to the game server as well.

MlgmXyysd (approved):

ghost (approved):

PR #128, opened 2022-04-22
Use count for level for equips (dev branch)

memetrollsXD:
Leveling up weapons is a pain in the ass. Instead of going through all of that pain, I've brought you a simple solution:

If the weapon is a weapon or artifact, use the count as level. For example:

- `!give 13501 90` will give you a lvl. 90 Staff of Homa


ghost (changes requested):
Weapons still have their ascension level (promoteLevel) that needs to be set accordingly when creating them at high levels. In addition, relics should be available to be created in bulk since they all have random stats. I recommend handling a 3rd argument for quantity.

KingRainbow44 (approved):

KingRainbow44 (approved):

ghost (approved):

PR #129, opened 2022-04-22
Use count for level for equips (stable branch)

memetrollsXD:
Leveling up weapons is a pain in the ass. Instead of going through all of that pain, I've brought you a simple solution:

If the weapon is a weapon or artifact, use the count as level. For example:

- `!give 13501 90` will give you a lvl. 90 Staff of Homa


ghost:
Please merge to dev branch first unless it is an important fix, thank you.

Issue #57, opened 2022-04-20
Upgrade to BEYOND Java 8


KingRainbow44:
At the moment Grasscutter is locked to Java 8. (if this isn't true, someone let me know)
The reasons for this are:
- The MongoDB mapper (at its current version) is not compatible with Java versions beyond 9.
- The client seems to not be able to connect to the **game server**

I've already attempted updating the MongoDB wrapper to 16 with success, but the second issue is still applicable.
Worst case scenario: the KCP server will need updating to support newer versions of Java.
Best case scenario: I screwed something up while trying to update.

If anyone wants the updated database wrapper, let me know!

EDIT:
- [ ] Refactor KCP system
- [ ] Refactor dispatch system
With both of these, Java 8+ should be supported just fine.

KingRainbow44:
Another good reason for doing this is because it seems people in the `#support` channel have a **ton** of issue with the Java version, since most have a newer version installed/don't know how to multiple versions of Java at the same time.

MlgmXyysd:
> * The MongoDB mapper (at its current version) is not compatible with Java versions beyond 9.

Can we migrated it to MySQL?

KingRainbow44:
> Can we migrated it to MySQL?

the idea isn't to migrate database (since the entire server literally RELYS on the mongo wrapper) and besides like I said, I already have the wrapper working on Java 16

4Benj:
> Can we migrated it to MySQL?

for the type of data we are saving MongoDB is more suitable in my opinion

KingRainbow44:
Finished with #130!

PR #130, opened 2022-04-22
Upgrade to versions of Java beyond 8

KingRainbow44:
**This update requires the use of JRE/JDK 11+.**
Completes task: #57

Included is:
- An upgrade to Morphia 2.x
- A keystore update (the existing one doesn't like newer versions of Java)


ghost (approved):

fumbling644o:
> A keystore update

...I don't see this?

MlgmXyysd:
I think we can implement https://github.com/Grasscutters/Grasscutter/pull/108 again

fumbling644o:
@MlgmXyysd #125 does the same thing, just without the fancy shorthand constructors from a newer version of Java.

But, we could, if we wanted to make the code slightly less unsightly.

MlgmXyysd:
> @MlgmXyysd #125 does the same thing, just without the fancy shorthand constructors from a newer version of Java.
> 
> But, we could, if we wanted to make the code slightly less unsightly.

Yes, I mean to change to new constructors

PR #139, opened 2022-04-23
Fix mitmproxy doesn't working

iTruth:
my mitmproxy is not installed in grasscutter root path, so I have to edit MITMDUMP_PATH in start_config.cmd  
in that situation mitmdump can't find proxy.py


MlgmXyysd (changes requested):
CR: will not work if `%MITMDUMP_PATH%` contains spaces

iTruth:
> CR: will not work if `%MITMDUMP_PATH%` contains spaces

Indeed! thanks for reminding  
I think the new commit should fix this  
let me know if something wrong

MlgmXyysd:
well, I made a better implementation #143

PR #157, opened 2022-04-23
Add another way to set talent level

omg-xtao:
Add another way to set current character talent level by using /talent command.

> /talent <n or e or q> <level> - Sets current character's talent level


alt3ri:
oh my thank you!

alt3ri (approved):

ghost (approved):

PR #167, opened 2022-04-24
Try to fix "No SSL cert found bind exception"

memetrollsXD:
Carefully guided by @fumbling644o on discord, trying to declutter support channel

### DIFF
--- a/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
+++ b/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
@@ -186,9 +186,9 @@ public final class DispatchServer {
 				} catch (Exception e) {
 					Grasscutter.getLogger().warn("[Dispatch] Unable to load keystore. Using default keystore password...");
 					KeyStore ks = KeyStore.getInstance("PKCS12");
-					ks.load(fis, "123456");
+					ks.load(fis, "123456".toCharArray());
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
-					kmf.init(ks, "123456");
+					kmf.init(ks, "123456".toCharArray());
 				} catch (Exception e) {
 					Grasscutter.getLogger().warn("[Dispatch] Error while loading keystore!");
 				}
--- a/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
+++ b/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
@@ -184,6 +184,7 @@ public final class DispatchServer {
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
 					kmf.init(ks, keystorePassword);
 				} catch (Exception e) {
+					Grasscutter.getLogger().warn("[Dispatch] Unable to load keystore. Using default keystore password...");
 					KeyStore ks = KeyStore.getInstance("PKCS12");
 					ks.load(fis, "123456");
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
--- a/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
+++ b/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
@@ -38,47 +38,48 @@ public final class DispatchServer {
 	private final InetSocketAddress address;
 	private final Gson gson;
 	private final String defaultServerName = "os_usa";
-	
+
 	public String regionListBase64;
 	public HashMap<String, RegionData> regions;
 
 	public DispatchServer() {
 		this.regions = new HashMap<String, RegionData>();
-		this.address = new InetSocketAddress(Grasscutter.getConfig().getDispatchOptions().Ip, Grasscutter.getConfig().getDispatchOptions().Port);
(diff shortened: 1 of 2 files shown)


src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
```diff
@@ -177,23 +177,23 @@ public void start() throws Exception {
...
-
+				KeyManagerFactory _kmf;
 				try {
 					KeyStore ks = KeyStore.getInstance("PKCS12");
 					ks.load(fis, keystorePassword);
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
+					_kmf = kmf;
 					kmf.init(ks, keystorePassword);
 				} catch (Exception e) {
 					Grasscutter.getLogger().warn("[Dispatch] Unable to load keystore. Using default keystore password...");
```
fumbling644o: `Using` -> `trying`

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
```diff
@@ -177,23 +177,23 @@ public void start() throws Exception {
...
 					KeyStore ks = KeyStore.getInstance("PKCS12");
 					ks.load(fis, "123456".toCharArray());
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
 					kmf.init(ks, "123456".toCharArray());
```
fumbling644o: Please consider adding another message notifying the user that the default password worked, and that they should probably change it in their config file.

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
```diff
@@ -177,23 +177,23 @@ public void start() throws Exception {
...
 				try {
 					KeyStore ks = KeyStore.getInstance("PKCS12");
 					ks.load(fis, keystorePassword);
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
+					_kmf = kmf;
 					kmf.init(ks, keystorePassword);
 				} catch (Exception e) {
 					Grasscutter.getLogger().warn("[Dispatch] Unable to load keystore. Using default keystore password...");
 					KeyStore ks = KeyStore.getInstance("PKCS12");
 					ks.load(fis, "123456".toCharArray());
```
fumbling644o: Move `"123456".toCharArray()` out so you don't make 2 allocations

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
```diff
@@ -177,23 +177,23 @@ public void start() throws Exception {
...
 					KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509");
 					kmf.init(ks, "123456".toCharArray());
-				} catch (Exception e) {
-					Grasscutter.getLogger().warn("[Dispatch] Error while loading keystore!");
```
fumbling644o: I think we still need this.

It's a bit ugly, but maybe a:

```java
catch (Exception originalEx) {
    try {
        // try to initialize kmf with "123456"
    }
    catch (Exception _) {
        Grasscutter.getLogger().warn("[Dispatch] Error while loading keystore!");
        originalEx.printStackTrace();
        throw originalEx; // don't care about the exception for the "123456" default password attempt
    }
}
```

fumbling644o (changes requested):

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
fumbling644o: (GitHub didn't load the last commit while I was reviewing, so pretend this is after line 192)

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
fumbling644o: (GitHub didn't load the last commit while I was reviewing, so pretend this is on line 189)

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
fumbling644o: (GitHub didn't load the last commit while I was reviewing, so pretend this is on line 188)

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
memetrollsXD: That will result in an error because Exception e was already catched before, and I don't know all of the error classes

src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
memetrollsXD: ??

memetrollsXD: Does using and trying make a different impact?

fumbling644o: No, it just makes the intent more clear to the user because it could still fail

ghost (approved):

Issue #188, opened 2022-04-24
BindException always occures in No-SSL env


Taskeren:
166行已经实例化了一个占用端口的服务器实例，然后在 SSL 错误抛出之后，182行再尝试新建服务器实例时必然会抛出 BindException

https://github.com/Grasscutters/Grasscutter/blob/3499b04e87d5969917e0038eaa24cdf121056265/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java#L166

https://github.com/Grasscutters/Grasscutter/blob/3499b04e87d5969917e0038eaa24cdf121056265/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java#L182

memetrollsXD:
Change your settings.json Keystore password to "123456"

Possibly fixed by #167

Taskeren:
> Change your settings.json Keystore password to "123456"
> 
> Possibly fixed by #167

改了之后并没有解决

memetrollsXD:
Is this with stable branch? Development branch is expected to have some issues

KingRainbow44:
this should be fixed in later versions of `development` so i will be closing this issue.

PR #189, opened 2022-04-24
bugfix when falling back to http server

Alan-Raddam:
If reading keystore failed, the origin line 176 has initialized a https server which occupies the port.  The new http server cannot be established. Even if we call httpsServer.stop(), it may takes nearly two minutes to wait for the https server to release the port. So I recommend initilize the https server after the keystore is correctly loaded. 

This is related to issue #188


memetrollsXD:
Possible duplicate of #167

Alan-Raddam:
Ahh... I think #167 didn't solve it. But obviously there is a merging conflict with #167. I will PR again after #167 is merged into the development branch.

Issue #142, opened 2022-04-23
release a Dockerflie？


ghost:
An amazing Job!
let's use docker to simplified the configuration process（╹◡╹）?

lihe07:
I don't think this environment is very troublesome to configure(just my personal opinion)
The external dependencies are not complicated

alphagocc:
Honestly, a docker image will be very convenient for many people to use

insoxin:
I'll get right to work on it.

ghost:
Now the project has too many issues about the environment. Like link error, java code error and so on.
It's a better way that run the server in a Linux docker, and then users only need to set the clients and redirect the traffic.

Problems about the server part can be reduced.

StoneMoe:
See pull request #255

Becods:
[HelloGrasscutter/HelloGrasscutter](https://github.com/HelloGrasscutter/HelloGrasscutter)

luaneko:
[Grassdocker][1] is my attempt to create a proper Docker image for Grasscutter. To quote the README:

> This repository provides a Dockerfile that can be used to build a completely self-contained Docker image for Grasscutter. It tracks the **development branch** and is confirmed to be working with Genshin 2.6.
>
> The resulting image is about 1.4 GB in size and has:
>
> - All required resources (bins, excels, etc.) baked into the image
> - Fixed reliquary excels with weight props
> - Proper image layering and multistaged build
> - Proper TLS root and server certificates with DNS names for `mihoyo.com`, `yuanshen.com` and `hoyoverse.com`
> - No unnecessary dependencies in the resulting image

[1]: https://github.com/chiyadev/Grassdocker

PR #212, opened 2022-04-25
Docker support

StoneMoe:
For issue #142

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
+ - Device: [e.g. iPhone6]
+ - OS: [e.g. iOS8.1]
+ - Browser [e.g. stock browser, safari]
+ - Version [e.g. 22]
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/custom.md
+++ b/.github/ISSUE_TEMPLATE/custom.md
@@ -1,0 +1,10 @@
+---
+name: Custom issue template
+about: Describe this issue template's purpose here.
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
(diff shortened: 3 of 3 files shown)


Mar0xy:
Why don't you just build the dockerfile and publish it as an image to docker hub so you only need docker-compose

StoneMoe:
> Why don't you just build the dockerfile and publish it as an image to docker hub so you only need docker-compose

I'm not the maintainer of this repository, let the right people to do "official release".
And I think it's kind of splitting the community.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM gradle:7.4.2-jdk8 AS builder
+WORKDIR /build
+COPY . .
+
+RUN tr -d '\015' <gradlew >gradlew-unix
+RUN chmod +x gradlew-unix
+RUN ./gradlew-unix jar
```
PinkD: how about merging them into one `RUN`

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: '3.9'
+
+services:
+  server:
+    image: grasscutter:1.0
+    build: .
+    volumes:
+      - ./resources:/app/resources
```
PinkD: considering exposing `config.json`

StoneMoe: > considering exposing `config.json`

Did you mean `<repo root>/config.json`?
Is this good to you? exposing everything might be configured
```
volumes:
  - ./resources:/app/resources
  - ./data:/app/data
  - ./config.json:/app/config.json
  - ./keystore.p12:/app/keystore.p12
```

but there's one more thing, docker will create a folder named `config.json` silently if `config.json` file not exist
which is really confused for someone don't create config file first
since `config.json` is in `.gitignore`

StoneMoe: And.. maybe a pre-created empty `resources` folder?

Dockerfile
StoneMoe: Done.

docker-compose.yml
memetrollsXD: resources folder gets created when grasscutter.jar gets run first time

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN tr -d '\015' <gradlew >gradlew-unix && \
+    chmod +x gradlew-unix && \
+    ./gradlew-unix jar
+
+
+FROM lwieske/java-8:server-jre-8u202-slim
+WORKDIR /app
+
+COPY --from=builder /build/grasscutter.jar grasscutter.jar
+
```
memetrollsXD: Run grasscutter.jar here so it auto-generates resources folder

memetrollsXD (changes requested):

docker-compose.yml
StoneMoe: > resources folder gets created when grasscutter.jar gets run first time

`resources` folder creation is not a big deal, docker will create it automatically.
`config.json` not existed at the beginning is the real problem, which will cause docker create `config.json` **folder**

Dockerfile
StoneMoe: ok, thanks.

docker-compose.yml
StoneMoe: Finally figure out a workaround

Usage:
```
docker-compose run config-gen
docker-compose up -d
```

StoneMoe:
Usage:
```
git clone https://github.com/Grasscutters/Grasscutter.git
cd Grasscutter
git submodule update --init --recursive

# ...prepare resources content

docker-compose run config-gen
docker-compose up -d
```

memetrollsXD:
Maybe add that to README too, might be useful

memetrollsXD (approved):

StoneMoe:
Here you go.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM gradle:7.4.2-jdk8 AS builder
+WORKDIR /build
+COPY . .
+
+RUN tr -d '\015' <gradlew >gradlew-unix && \
+    chmod +x gradlew-unix && \
+    ./gradlew-unix jar
+
+
+FROM lwieske/java-8:server-jre-8u202-slim
```
Becods: I think it would be better to use `openjdk:8-jre-slim`, at least it's official release.

I know developers recommend using 8u202 because it's the last free version, and although the official image of openjdk doesn't release a Linux-based docker version of java 8u202, we can use `azul/zulu-openjdk:8u202` or other java distributions.

StoneMoe: I do heard about openjdk alternative, and java license changes...

I just followed README instruction, since I'm not really familiar with Java.
and really don't know why we can't just use newest OpenJDK :P

Anyway, I'll do as you say :D

ghost (approved):

PR #235, opened 2022-04-25
add auto spawn entity

WetABQ:
Read entities config from YSLua scene data, and auto spawn those entities.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+[{*.java,*.kt,*.kts}]
+insert_final_newline = true
+indent_style = tab
+indent_size = 4
+max_line_length = 120
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,8 @@ plugins {
 
     // Apply the application plugin to add support for building a CLI application
     id 'application'
+    id 'org.jetbrains.kotlin.jvm' version '1.6.21'
+    id 'org.jetbrains.kotlin.plugin.serialization' version '1.6.20'
 }
 
 sourceCompatibility = 16
@@ -23,7 +25,7 @@ repositories {
 
 dependencies {
     implementation fileTree(dir: 'lib', include: ['*.jar'])
-    
+
 	implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.32'
 	implementation group: 'ch.qos.logback', name: 'logback-core', version: '1.2.9'
     implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.9'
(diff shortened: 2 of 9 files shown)


ghost:
We cannot include the yslua files because of takedown risk.

WetABQ:
> We cannot include the yslua files because of takedown risk.

could we include yslua just like `BinOutput`?

I do not upload on this pr. i think we could upload those files on another repo, just like `GenshinData`

ghost:
The thing is that the lua files get taken down themselves. See [https://github.com/TheLostTree/YSLua](https://github.com/TheLostTree/YSLua), if you add a link to any lua files, hoyo will find a way to go after them.

WetABQ:
> The thing is that the lua files get taken down themselves. See https://github.com/TheLostTree/YSLua, if you add a link to any lua files, hoyo will find a way to go after them.

Could we add an option to this feature and add those files by users themselves?

If they don't have any YSLua in the resources folder, this feature will auto turn off

ghost:
I'd rather not tell users do go find things that would cause a dmca.

Mar0xy:
I do see the appeal in something like this considering without entities it does feel really empty. and something like this normally should be part of a private server

ghost:
You can do them in other ways without telling users to include luas.

WetABQ:
> You can do them in other ways without telling users to include luas.

It currently doesn't include any Lua, only the json files

Colerar:
The code style of `.kt` is inconsistent with current `.java` files. (For example, `.java` use 2 tab indent, `.kt` use 4 spaces)

We'd better add a `.editorconfig` to make code style uniform.

memetrollsXD:
@Melledy The data itself doesn’t get taken down you say? I’m assuming he only means the JSON files, so that would also be data

TheLostTree:
Personally, i think it's totally fine. Reupload the lua/jsons on a different repo just like the protobufs +  make it disable itself if the folder isnt there.

WetABQ:
> 

im doing serializer for it right now, thanks for suggestion

ghost:
> Personally, i think it's totally fine. Reupload the lua/jsons on a different repo just like the protobufs + make it disable itself if the folder isnt there.

The problem is any sites or repos hosting the luas would receive dmca notices, i think its just safer to parse to json for the time being.

ghost:
@WetABQ Can you merge with to the dev-entity branch instead of development?

WetABQ:
> @WetABQ Can you merge with to the dev-entity branch instead of development?

yes I can

WetABQ:
> > Personally, i think it's totally fine. Reupload the lua/jsons on a different repo just like the protobufs + make it disable itself if the folder isnt there.
> 
> The problem is any sites or repos hosting the luas would receive dmca notices, i think its just safer to parse to json for the time being.

I initially wanted to package all the data into a single JSON just like @Melledy has done in `dev-branch`, but I found that to be too much performance overhead. The current `SceneDataLoader` is implemented via lazy delegation for faster data reading. So I will keep the original JSON file structure.

PR #248, opened 2022-04-26
Add `bless of moon` feature

Konoyoru435:
In a nutshell: Item ID 1202 (Moon Card) is now available, adding support for the `bless of moon` feature.

Working list and main ideas:
- [x] Records the card activation time, duration, and activation status, and returns rewards based on these parameters;
    - Given before the player enters the world;
    - Activate the status after using the item and increase the duration.
- [x] Only one card can be used at a time;
- [ ] If a player with an active moon card waits in the game until the next day, then the server should also send packets. (Very rough code)
    - [ ] Need to separate `TimerManager`.

Protos:
- `CardProductRewardNotify.proto`

PR Here: https://github.com/Grasscutters/Grasscutter-Protos/pull/10

Testing Work:
Failure


Yazawazi:
Can I use `quartz`?

memetrollsXD:
quartz? :thonk:

Yazawazi:
> quartz? :thonk:

http://www.quartz-scheduler.org/

memetrollsXD:
I think that works, maybe even a little bit of overkill lmao

Yazawazi:
Haha, my idea is to use `Quartz` to implement a `TimerManager`.For example, a task sends rewards to moon-card-activated players at 4 am (UTC+8) every day. Maybe we can use it in the future. I'm not sure, let's try it out.

Issue #148, opened 2022-04-23
Exception reflections.ReflectionsException: Scanner TypeAnnotationsScanner was not configured


StillMi:
Exception in thread "main" org.reflections.ReflectionsException: Scanner TypeAnnotationsScanner was not configured
        at org.reflections.Store.get(Store.java:39)
        at org.reflections.Store.get(Store.java:61)
        at org.reflections.Store.get(Store.java:46)
        at org.reflections.Reflections.getTypesAnnotatedWith(Reflections.java:429)
        at org.reflections.Reflections.getTypesAnnotatedWith(Reflections.java:416)
        at emu.grasscutter.command.CommandMap.scan(CommandMap.java:147)
        at emu.grasscutter.command.CommandMap.<init>(CommandMap.java:19)
        at emu.grasscutter.server.game.GameServer.<init>(GameServer.java:60)
        at emu.grasscutter.Grasscutter.main(Grasscutter.java:80)
When it starts
Cannot continue

ywnkm:
I solved this problem by updating the version of reflections
```
implementation group: 'org.reflections', name: 'reflections', version: '0.10.2'
```

StillMi:
thank you  problem solved

KingRainbow44:
was this issue on dev or stable?

czy20010112:
> I solved this problem by updating the version of reflections
> 
> ```
> implementation group: 'org.reflections', name: 'reflections', version: '0.10.2'
> ```

sorry to trouble but could u please tell me how to update this? do i need to download the source code and change it by myself? or could u give me a command to update it? thanks a lot 
java version "18.0.1" 2022-04-19  centos 7.9 mongodb 4.4
Java(TM) SE Runtime Environment (build 18.0.1+10-24)
Java HotSpot(TM) 64-Bit Server VM (build 18.0.1+10-24, mixed mode, sharing)

PR #249, opened 2022-04-26
update org.reflections

F-Unction:
0.9.12->0.10.2
to avoid `[INFO] Unknown command` (#148)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ dependencies {
 	implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.8'
 	implementation group: 'com.google.protobuf', name: 'protobuf-java', version: '3.18.1'
 	
-	implementation group: 'org.reflections', name: 'reflections', version: '0.9.12'
+	implementation group: 'org.reflections', name: 'reflections', version: '0.10.2'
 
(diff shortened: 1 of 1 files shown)


memetrollsXD (approved):

KingRainbow44 (approved):

KingRainbow44:
does the reflector have to be updated with it?

F-Unction:
no, everything works well.

Konoyoru435 (approved):

1shin-7:
See #251

4Benj:
> See #251

I think the problem with #251's solution is that you cannot use the Command annotation with plugins when they get implemented fully since it requires specifying a specific package

F-Unction:
so, should I use the solution of #251 ? this is a pr to stable branch

exzork:
dev already use 0.10.2, so this pr shouldn't affect dev
https://github.com/Grasscutters/Grasscutter/blob/7739fc1aeb36989d6a452d22a379bdc5db477f60/build.gradle#L35

F-Unction:
yeah, this pr is to solve #148 on stable branch, so i think change the version of org.reflection is enough

ghost (approved):

PR #255, opened 2022-04-26
Docker support

### LINKED ISSUE #142: release a Dockerflie？

StoneMoe:
For issue https://github.com/Grasscutters/Grasscutter/issues/142
Move from https://github.com/Grasscutters/Grasscutter/pull/212


ossima-ze:
Hi. thanks for your jobs. It helps a lot, btw I'm using your fork [StoneMoe/Grasscutter], (<https://github.com/StoneMoe/Grasscutter>). But it cannot work correctly.

Here's the errors

## docker-compose version error

I have a question, you specified `version: 3.9` in `docker-compose.yaml`, and technicality nothing seems wrong according to [this post](<https://forums.docker.com/t/docker-compose-version-3-8-or-3-9-for-latest/102439>). But error raised in my docker, that's why? Here's the error info:

```
ERROR: Version in ~'./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
```

Is there any demand to specified v3.9? Replacing `version: 3.9` with `version: 3`  won't raise version not supported error anymore. BTW,  I'm using docker-compose verison 1.25.0, the newest is version 1.29.0, things seems to be ok with my other docker with compose v3

Why `version: '3.9'` is specified? Currently the max version of compose is 3.8, seeing the [doc](<https://docs.docker.com/compose/compose-file/compose-versioning/>). Is is for `profiles` property？

## donotstart property error

And `profiles: - donotstart` caused syntax error in my docker, but `donotstart` cannot be searched in [docker compose doc](<https://docs.docker.com/search/?q=donotstart>). Seeing this [issue](<https://github.com/docker/compose/issues/1439>) on Docker/compose, I realise that you might want this service don't start automaticly. But this also raise error. 

Ignoring `profiles: - donotstart` options and replaced it with `restart: no` can fix this error, and things went fine.

## another error at final

With specified `version: 3` and replaced `profils: donotstart` with `restart: no`, I can start to build the server finally, but, error happened:

```
ERROR: Service `server` failed to build: COPY failed, file not found in build context or excluded by .dockerignore: stat Grasscutter-Protos/proto/: file does not exist
```

Becods:
> With specified version: 3 and replaced profils: donotstart with restart: no, I can start to build the server finally, but, error happened:
> ```
> ERROR: Service `server` failed to build: COPY failed, file not found in build context or excluded by .dockerignore: stat Grasscutter-Protos/proto/: file does not exist
> ```

I think this is an oversight that was created when writing the yml.

You need update submodule.
```
git submodule update --init --recursive
```

And later I will try to pull a tested version.

StoneMoe:
> I have a question, you specified `version: 3.9` in `docker-compose.yaml`, and technicality nothing seems wrong according to [this post](https://forums.docker.com/t/docker-compose-version-3-8-or-3-9-for-latest/102439). But error raised in my docker, that's why? Here's the error info:
> ERROR: Version in ~'./docker-compose.yml" is unsupported.

when I wrote the compose file I assume user will install the latest docker version.
I'll try to support older compose version, but can't give you a guarantee
and still highly recommended to upgrade to latest docker version :D 

> Why `version: '3.9'` is specified? Currently the max version of compose is 3.8, seeing the [doc](https://docs.docker.com/compose/compose-file/compose-versioning/). Is is for `profiles` property？

latest version is 3.9, docker documentation is inconsistent with implementation
and YES, it's for profiles property.

> And `profiles: - donotstart` caused syntax error in my docker, but `donotstart` cannot be searched in [docker compose doc](https://docs.docker.com/search/?q=donotstart). Seeing this [issue](https://github.com/docker/compose/issues/1439) on Docker/compose, I realise that you might want this service don't start automaticly. But this also raise error.

`donotstart` is just a custom name, indicate these service should be started specifically
For our use case, `res-gen` `config-gen` only start once for generating files, manually.
see https://docs.docker.com/compose/profiles/#compose-documentation

> Ignoring `profiles: - donotstart` options and replaced it with `restart: no` can fix this error, and things went fine.

`restart: no` only affect the behavior when container exit, so not the same thing.

> With specified `version: 3` and replaced `profils: donotstart` with `restart: no`, I can start to build the server finally, but, error happened:
> 
> ```
> ERROR: Service `server` failed to build: COPY failed, file not found in build context or excluded by .dockerignore: stat Grasscutter-Protos/proto/: file does not exist
> ```

you can follow docker setup instruction in README to fix this.
`docker-compose run res-gen` will update submodule for you
or you can run `git submodule update --init --recursive` by yourself

ossima-ze:
@StoneMoe ohh, `git submodule update --init --recursive` works for me. Docker-compose now can be deployed, but something still wrong, when I enter `docker logs grasscutter_server_1 -f `, two errors happened:

image

Config cannot be saved, and ExcelBinOutput error for some reasons, this is what I meet when I deployed Grasscutter without docker, following the doc.

Of course you can run the docker, but what happens with mine docker! That's so sad.......

Oh, and how to interact  with the Grasscutter server? I don't know how to add a new user, or I should use `docker-compose run` instead of `docker-compose up -d`?

Becods:
> Oh, and how to interact with the Grasscutter server? I don't know how to add a new user, or I should use `docker-compose run` instead of `docker-compose up -d`?

```
docker attach grasscutter_server_1
```

StoneMoe:
> Config cannot be saved

Remove `config.json` **folder** if exists, and rerun `docker-compose run config-gen`

Becods:
Perhaps there is a more elegant solution to this, I think you could create an entrypoint file that causes the generated configuration file and updated data to run at startup.

I'll probably provide an example later on.

StoneMoe:
> Perhaps there is a more elegant solution to this, I think you could create an entrypoint file that causes the generated configuration file and updated data to run at startup.
> 
> I'll probably provide an example later on.

I would prefer leave all these logic in docker compose file, to do it in a manually way, since resource management is still not exists in this project

ossima-ze:
> > Config cannot be saved
> 
> Remove `config.json` **folder** if exists, and rerun `docker-compose run config-gen`

@StoneMoe Here, with `docker-compose run config-gen` executed, things seems will runs normally.

But with your new release, `grasscutter_server_1` will automaticly exit, it's not attachable. Only `grasscutter_mongodb_1` survive.

image

After executing `docker-compose run config-gen`, I can see `config-gen` will be started but it will exit after seconds after executing `docker-compose up -d`, same as `grasscutter_server_1`, I don't think it runs as expectations.

StoneMoe:
> But with your new release, `grasscutter_server_1` will automaticly exit, it's not attachable. Only `grasscutter_mongodb_1` survive.

log please

> After executing `docker-compose run config-gen`, I can see `config-gen` will be started but it will exit after seconds after executing `docker-compose up -d`, same as `grasscutter_server_1`, I don't think it runs as expectations.

service `config-gen` is for generating config file, it should exit after generating `config.json`

ossima-ze:
I got it, the reason why I meet this error is that I didn't execute `res-gen` first.

So. I'm supposed to run `res-gen` and `config-gen`, and all work fine, except one error mentioned above.

```
[ERROR] Relic properties are missing weights! Please check your ReliquaryMainPropExcelConfigData or ReliquaryAffixExcelConfigData files in your ExcelBinOutput folder.
```

No need to worry about it? Cause I'm confused to how to connect to the server, so I don't known the server works normally whether or not (The wiki do not help at all, some 3rd. documents alse make me confused).

StoneMoe:
> [ERROR] Relic properties are missing weights! Please check your ReliquaryMainPropExcelConfigData or ReliquaryAffixExcelConfigData files in your ExcelBinOutput folder.
> ```
> 
> No need to worry about it? Cause I'm confused to how to connect to the server, so I don't known the server works normally whether or not (The wiki do not help at all, some 3rd. documents alse make me confused).


I believe it's a known issue which you'd also encounter by following wiki to build your own `resources` folder.

ossima-ze:
Opps! I had tried again to make my server run. Currently running Grasscutter on WSL, but failed to join the game.

Here's my docker's logs for `grasscutter_server_1`

```
~/opensource/docker/Grasscutter (docker ✔) docker attach grasscutter_server_1
[17:17:03] [INFO] Game Server started on port 22102
[17:17:12] [INFO] [Dispatch] Client 172.21.0.1/172.21.0.1:35004 is trying to log in
[17:17:12] [INFO] [Dispatch] Client 172.21.0.1/172.21.0.1:35004 logged in as 10002
[17:17:15] [INFO] [Dispatch] Client 172.21.0.1/172.21.0.1:35012 succeed to exchange combo token
[17:17:23] [INFO] Client 172.21.0.1/172.21.0.1:35040 request: query_cur_region_os_usa
```

Yes I can login with my custom account name, but then the game says I have disconnect

image

Can you tell me why? Seaching the google and baidu, I still cannot get any solutions. 

---

And another question is how do you proxy the server? The ports that all you exposed seems only for mongodb, proxying certs and others. I run a proxy server on Windows, the command is `mitmdump -s ./proxy.py -k`

image

The last of the last, THANK YOU sincerely!

StoneMoe:
> Currently running Grasscutter on WSL, but failed to join the game
> Yes I can login with my custom account name, but then the game says I have disconnect

most likely UDP port is not reachable, I ran docker desktop on wsl2 and everything works fine.

> And another question is how do you proxy the server? The ports that all you exposed seems only for mongodb, proxying certs and others. I run a proxy server on Windows, the command is `mitmdump -s ./proxy.py -k`
> The last of the last, THANK YOU sincerely!

`network_mode: service:mongodb` means use the same network stack as mongodb service

ossima-ze:
> most likely UDP port is not reachable, I ran docker desktop on wsl2 and everything works fine.

But I also ran server on my WSL2, so the question becomes how do you set up your Windows proxy? Without running mitmdump, I tried to connect to `127.0.0.1` on all the ports were exposed (80, 443, 8888, 22102, `27017`, 22102) on Windows proxy settings, they all not work. Which port that your Windows proxy connected to?

StoneMoe:
> Which port that your Windows proxy connected to?

Use the port mitmproxy listening on, you can read https://en.wikipedia.org/wiki/Man-in-the-middle_attack to understand how mitm works.

and use discord for user support please.

PengCheong:
I got it, so to run the server, my VPS still need a backend service (like mitmproxy) to listen at port 8080 (or other). Thank you again.

ossima-ze:
Oh, I got it, so to run the server on my VPS, another backend service for listening at port 8080 and others is needed. I can deploy a mitmproxy docker image as well. Thank you!

memetrollsXD:
Does this PR still require attention? Else I will close this **without merging**

StoneMoe:
> Does this PR still require attention? Else I will close this **without merging**

i don't understand, is there anything wrong with this PR?

StoneMoe:
I think this PR is bug free and ready to be reviewed.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+
+FROM azul/zulu-openjdk:8u202
+WORKDIR /app
+
+COPY --from=builder /build/grasscutter.jar grasscutter.jar
+
+# mount these folder/file at runtime please
+# resources/ data/ config.json keystore.p12
+COPY Grasscutter-Protos/proto/ ./proto
```
memetrollsXD: Protos are already included and their classes are now automatically built

image

memetrollsXD (changes requested):

66Leo66:
Why do we need this?
I mean, why not use a JRE container, attach volume containing server data, and run the server?

ossima-ze:
> Why do we need this? I mean, why not use a JRE container, attach volume containing server data, and run the server?

So, how do you run mongodb inside jre container? I suppose mongodb / mitmdump and the server itself inside docker is much more convenient for most people.

AdemJensen:
> > Why do we need this? I mean, why not use a JRE container, attach volume containing server data, and run the server?
> 
> 
> 
> So, how do you run mongodb inside jre container? I suppose mongodb / mitmdump and the server itself inside docker is much more convenient for most people.

Why do we need a mitmdump on server side...
Unfortunately this design is a little bit awkward, the pros will simply keep Grusscutter and Mongo in separate containers, while the beginners will just deploy them manually...

66Leo66:
> > Why do we need this? I mean, why not use a JRE container, attach volume containing server data, and run the server?
> 
> So, how do you run mongodb inside jre container? I suppose mongodb / mitmdump and the server itself inside docker is much more convenient for most people.

Personally speaking, I'm against this idea. My own deployment uses docker compose, running mongodb and grasscutter in two containers. And I also put mitm in a seperate compose file because I don't need proxy on my VPS where grasscutter runs.

In conclusion, all in one is convenient for quick start but much harder to customize.

docker-compose.yml
```diff
@@ -0,0 +1,83 @@
...
+      git clone --depth 1 https://github.com/Dimbreath/GenshinData.git /tmp/GenshinData;
+      git clone --depth 1 https://github.com/radioegor146/gi-bin-output.git /tmp/gi-bin-output;
+
+      echo Copying resources...;
+      mkdir -p /app/src/resources/;
+      mv /tmp/GenshinData/TextMap /app/src/resources;
+      mv /tmp/GenshinData/Subtitle /app/src/resources;
+      mv /tmp/GenshinData/Readable /app/src/resources;
+      mv /tmp/GenshinData/ExcelBinOutput /app/src/resources;
+      mv /tmp/gi-bin-output/2.5.52/Data/_BinOutput /app/src/resources/BinOutput;
```
lilmayofuksu: Since radioegor146's repo got DMCA'd, you need to use different data from somewhere else.

StoneMoe: Thanks, fixed.

Dockerfile
StoneMoe: done.

memetrollsXD:
Waiting for development to merge to stable

memetrollsXD (approved):

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+    entrypoint: ["/bin/sh", "-c", "
+      echo [!!] resources folder will be deleted;
+      echo - Any key to continue;
+      echo - Ctrl+C  to abort;
+      read answer;
+
+      echo Deleting old resources...;
+      rm -rf /app/src/resources;
+      
+      echo Downloading resources...;
```
KingRainbow44, line 64: consider adding yslua or using `Grasscutter_Resources`

KingRainbow44 (changes requested):

KingRainbow44:
> would rather have external repos handle the docker stuff

@Melledy

PR #262, opened 2022-04-26
load all classes from plugin

exzork:
PluginManager only load main class from plugin file and it can trigger NoClassDefFoundError, so i'm opening this PR to load all classes from plugin file.


KingRainbow44 (approved):
thank you very much, saves me a few hrs of stackoverflowing lmao

PR #265, opened 2022-04-26
Update Grasscutter.java to fix pr251

Secretboy-SMR:


KingRainbow44:
the reflector isn't just for the command system, it can be used in other places too. if the server is unable to load commands then there might be another issue besides the reflector.

PR #268, opened 2022-04-26
Add a developer-friendly API to Grasscutter

KingRainbow44:
Adds hooks to certain classes allowing basic (and frequent) actions to be completed in a simple manner.

Implements:
- `PlayerMailReceiveEvent` - Called when the player receives mail
- `PlayerQuitEvent` - Called when the player leaves the server
- Plugin data folder (located in `/plugins/(plugin name)/`)
- Obtaining resources from the plugin's JAR file
- Overhauled event handler system
- Classloader for other files in plugins
- A "server hook" allowing plugins to do server-related actions 10x faster
- `PlayerJoinEvent` - Called when the player joins the server
- `PlayerCreationEvent` - Called before the player instance is created

**For Developers**:
This PR can be accessed through maven using `tech.xigam.grasscutter:1.0.3-dev`


ghost (approved):

src/main/java/emu/grasscutter/server/packet/recv/HandlerSetPlayerBornDataReq.java
```diff
@@ -91,6 +92,10 @@ public void handle(GameSession session, byte[] header, byte[] payload) throws Ex
 			Grasscutter.getLogger().error("Error creating player object: ", e);
 			session.close();
 		}
-	}
 
+		// Call join event.
+		PlayerJoinEvent event = new PlayerJoinEvent(player); event.call();
```
ghost: SetPlayerBornReq is when players are created (right after opening cutscene) not when they join, if you want to do the PlayerJoinEvent, when they actually join, it should at the end of Player::onLogin

memetrollsXD (approved):
Everything looks fine except for the change that Melledy proposed

PR #226, opened 2022-04-25
[BREAKING CHANGE] proto auto compiled by gradle

WetABQ:
move proto to git submodule
the gradle will auto compile all proto files in ./proto/ and generate code in src/generated/

fix all the bugs produced in this migration process, I've tested in-game, and it seems no big issues.

proto update Grasscutters/Grasscutter-Protos/pull/9

PR #269, opened 2022-04-26
Implement some features simply

Kengxxiao:
1. Add commands /coop and /tpall.
    - /coop is used to force someone to teleport into others' world forcely
    - /tpall is used to teleport all players in your world to your position
2. Implement simple monster drop and shop which be controlled by configuration file.
    - There're 2 issues, first one is when player picked up item in MP world the entity didn't disappear in others' client, and the second one is ```killEntity``` method could receive invalid attacker (which caused ```callDrop``` not be invoked, because ```EntityItem``` requires ```GenshinPlayer``` instance to be constructed) , I don't know how to fix them sorry :(
3. Implement lunch box but players only can eat Delicious Sweet Madame.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ hs_err_pid*
 build/
 out/
 
+# Ignore Gradle properties
+gradle.properties
+
 # Eclipse
 .project
 .classpath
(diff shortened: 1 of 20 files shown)


WetABQ:
after #226, u should del those `Outer` classes, which are generated by `protoc`
try to pull and merge, to resolve merge conflicts of u pr, thanks very much

memetrollsXD:
Please make a separate pull request per feature, we can't manage them all in one.. if there are known issues you shouldn't also be pushing just yet

PR #272, opened 2022-04-27
Fix "Unknown Command"

Lynricsy:
Add reflection-0.10.2.jar to lib.


ghost:
We already have reflections in gradle build?

PR #281, opened 2022-04-27
Maybe make the teleport command better

QiE2035:
1. Maybe make the "teleport" command better
2. An command "translate" to translate the player forward to distance.


QiE2035:
Sorry. I use the wrong api.

Issue #287, opened 2022-04-27
Add wildcard characters to permission nodes under the same namespace


lhhxxxxx:
Using wildcards to match all permission nodes under the same namespace will make it easier to grant and remove permissions. For example, we can use player.* or player to match all permission nodes under the player namespace.

Perhaps we could use the following code as an alternative.

`
public boolean hasPermission(String permission) {
	return this.permissions.contains(permission) || this.permissions.contains("*") || (this.permissions.contains("player") || this.permissions.contains("player.*")) && permission.startsWith("player.") || (this.permissions.contains("server") || this.permissions.contains("server.*")) && permission.startsWith("server.") ? true : false;
}
`

lhhxxxxx:
#288

PR #288, opened 2022-04-27
Update Account.hasPermission()

lhhxxxxx:
Add wildcard characters to permission nodes under the same namespace. (simple implementation)

### DIFF
--- a/src/main/java/emu/grasscutter/game/Account.java
+++ b/src/main/java/emu/grasscutter/game/Account.java
@@ -104,7 +104,10 @@ public class Account {
 	}
 
 	public boolean hasPermission(String permission) {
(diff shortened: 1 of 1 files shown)


lhhxxxxx:
#287

4Benj:
The problem with this solution is that there could be more namespaces in the future and custom permissions added through plugins.

lhhxxxxx:
> The problem with this solution is that there could be more namespaces in the future and custom permissions added through plugins.

I think to solve this problem fundamentally, we have to change the storage of permission nodes from `String` with '.' in it to something like `List<String>` , otherwise the complexity of processing would be ridiculous.

(So this pr is a temporary solution to make it easier to grant or remove permissions on the current server until the overall refactoring of the permissions node storage)

ghost (approved):

PR #292, opened 2022-04-27
Implement simple drop, shop and lunch box features

Kengxxiao:
I fixed my code after #226.

- Implement simple drop system, it can be controlled in ```Drop.json``` in ```data``` folder.
    - There're some changes about ```EntityItem```. In official server dropped items can be shared (regional specialty) or independent(other normal monster material), I think they should be separated so I changed code to make server send broadcast when player picked **shared** item or send single packet to player only when it picked **independent** item.
    - ```EntityItem``` is **shared** by default now, so ```/drop``` command will summon a **shared** item.
    - It fixed my issues in #269 , because shared items will disappear in other players' vision when someone picked it up, and independent item only could be seen by single player.
- Implement simple shop system, it can be controlled in ```Shop.json``` in ```data``` folder.
    - Only mora cost is supported.
- Implement lunch box and widget equip.

To @memetrollsXD in #269 : 
Thanks for your reply but I cannot split my code for every features because its coupled with base game object sry :(


alphagocc:
`Thanks for your reply but I cannot split my code for every features because its coupled with base game object sry :(`
I think decoupling modules is contributors’ duty. 
I don't think there's a strong relation between Drop, Shop and Lunch box system.
At least, you need to split 1 commit to 3 commits.
And I also find that you tweak the gacha system and revert it which is unable to infer from the commit message.

PR #293, opened 2022-04-27
Add command /coop and /tpall

Kengxxiao:
Add 2 commands

1. ```/coop <player id> <target player id>```
    - Forces someone to join the world of others‘ world
2. ```/tpall```
    - Teleports all players in your world to your position

I made them to test some changes in MP mode.


memetrollsXD:
Commands work, but co-op is buggy, and we cannot compare it to normal co-op at this moment

memetrollsXD (approved):

PR #297, opened 2022-04-27
Implement simple monster drop

Kengxxiao:
It implemented simple monster drop feature which be controlled by configuretion file ```Drop.json```, and it fixed a issue when someone picked items in MP mode.

More details in #292

In the preview, the dropped reliquary is defined as ```shared```, so it disappeared when ganyu picked it up.
Another intertwined fate is defined as ```independent``` so it's not affected in MP mode.

{_AGPZCR{N6US)U78VSRGO2
X`(3B(FGPQ2}D(AY~M0(D2I


memetrollsXD (approved):

memetrollsXD:
Tested and working

WetABQ:
why not implement it with `MonsterExcelConfigData.json`, and I think the drop data is in the key called "HpDrops" and also with the drop probability of items.

and GC even has done the deserialization of this json in `emu.grasscutter.data.def.MonsterData`

PR #331, opened 2022-04-28
Add luaj and parse lua files & bind lua funtions in java & fix errors in build.gradle & make proto as a subproject

RimuruChan:
Add luaj and parse lua files
Bind lua funtions in java
Fix task jar pack proto files into jar
Make proto as a subproject


QiE2035:
I have stupid question: where can find these lua files?

ghost:
@RimuruChan I'm working on my implementation already, will be pushing mine in a bit.

PR #336, opened 2022-04-28
Added support for riding Waveriders

lilmayofuksu:
Hardcoding waverider id is not pretty, but it does work. I also added [VehicleStaminaNotify.proto](https://github.com/Grasscutters/Grasscutter/compare/development...lilmayofuksu:development-vehicle?expand=1#diff-a42376da880410a95f6662e60d4495135301c334d4f7a7d09a22c9ad96c1413f) but it is not used for now. 

Edit: There isn't a limit related to how much you can spawn waveriders, so that probably needs to fixed.

### DIFF
--- a/proto/VehicleInteractReq.proto
+++ b/proto/VehicleInteractReq.proto
@@ -1,0 +1,11 @@
+syntax = "proto3";
+
+option java_package = "emu.grasscutter.net.proto";
+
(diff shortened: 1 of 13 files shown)


memetrollsXD (approved):
Code looks good, tested on co-op and worked

PR #340, opened 2022-04-28
Limit the amount of vehicles that a player can spawn

lilmayofuksu:
This is not tested(yet). It *should* work but I honestly don't know.


lilmayofuksu:
I fixed the ConcurrentModificationException, server now ejects the boat members if there are any, then it kills the existing boat.
Tested with singleplayer

Edit: Tested coop, it also works fine. Thank you to everyone (especially @alt3ri) for testing with me!

memetrollsXD (approved):

PR #351, opened 2022-04-29
Add Spotless gradle plugin for enforcing a code format

2bllw8:
Hi everyone! First time contributing to this project.

While looking at the codebase I noticed that since there are many people contributing there are a lot of inconsistencies with files format (e.g. tabs vs 4-space vs 8-space...), and this also leads to additional edits in many other pull requests because people have different habits.

I added a gradle plugin called ["spotless"](https://github.com/diffplug/spotless) that can enforce a code format and also automatically apply it (with the `:spotlessApply` gradle target).

I configured and executed it with one of the default available styles, but I'm willing to tweak its settings so that a good styling that everyone agrees on.

Finally, I added a new CI script that checks for code format in pull requests.

Let me know what do you think of this and what would a good code styling choice be for the project in your opinion.

Note: before merging, it is needed to rebase and execute the spotlessApply target again so it's up-to-date


PinkD:
https://github.com/Grasscutters/Grasscutter/blob/553f4df8d6b599fba0d2c3558a95089a2e771a93/build.gradle#L71-L79

Can't believe [google uses 2-space indent](https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation) 🤣 
Maybe we should create a poll to decide which one we should choose?
I prefer 4-space btw

2bllw8:
I uploaded a new config that uses 4-spaces indent so you can see how it looks like.

If you want to ping me to discuss about this patch, I'm in your Discord server (same username)

KingRainbow44 (changes requested):
Isn't `gradlew` not meant to be empty?

2bllw8:
> Isn't `gradlew` not meant to be empty?

It is not empty, the diff is empty because the text is unchanged. The first commit just sets the execute bit (see the change `100644 → 100755` near the file name on GitHub) so that Linux and MacOS users can invoke it by typing `./gradlew` instead of explicitly invoking an interpreter like `bash` to run it.

KingRainbow44 (approved):

KingRainbow44:
image
if its possible to ignore some formatting with comments, let me know and I'll reopen this PR

2bllw8:
@KingRainbow44 yes it is possible to disable the formatter using some comments, like this:

```java
// @formatter:off
PlayerCreation event = new
   PlayerCreationEvent(
 session,
    Player.class);
// @formatter:on
```

PR #353, opened 2022-04-29
Support Teleport from console & custom server avatar

omg-xtao:
eg: `tp @10001 x y z 3`


ghost (approved):

mikuyourworld:
I think you can also add custome server's name...

alt3ri:
> I think you can also add custome server's name...

you can change server name in config.json

mikuyourworld:
> > I think you can also add custome server's name...
> 
> you can change server name in config.json

The name in node gameserver?
I do not mean this name.

I mean the nickanme who sent message to you when you login or you use commands.. the nickname is Server now.

alt3ri:
you can find it in game/packet/send/

mikuyourworld:
> you can find it in game/packet/send/

Yes, I saw that..
So I am thinking that he added costome avatar, maybe he can add custome nickname together, although if I want I can change the nickname in the source.

omg-xtao:
> > you can find it in game/packet/send/
> 
> Yes, I saw that.. So I am thinking that he added costome avatar, maybe he can add custome nickname together, although if I want I can change the nickname in the source.

Done.

src/main/java/emu/grasscutter/Config.java
```diff
@@ -51,6 +51,7 @@ public static class RegionInfo {
 	
 	public static class GameServerOptions {
 		public String Name = "Test";
+		public String NickName = "Server";
```
memetrollsXD: Specify more, like "serverNickname"

src/main/java/emu/grasscutter/command/commands/TeleportCommand.java
```diff
@@ -1,67 +1,88 @@
...
     public void execute(Player sender, List<String> args) {
-        if (sender == null) {
-            CommandHandler.sendMessage(null, "Run this command in-game.");
+        int target;
+        if (args.size() < (sender == null ? 4 : 3)) {
+            CommandHandler.sendMessage(sender, sender == null ? "Usage: /tp @<player id> <x> <y> <z> [scene id]" :
+                    "Usage: /tp [@<player id>] <x> <y> <z> [scene id]");
             return;
         }
+        if (args.get(0).startsWith("@")) {
```
memetrollsXD, line 23: Stay consistent with other commands and dont use @, but just the UID

memetrollsXD (changes requested):

src/main/java/emu/grasscutter/command/commands/TeleportCommand.java
omg-xtao: How do I handle scene id and uid if I don't use the @ prefix?

omg-xtao: All the parameters are numbers, so when you enter four parameters, how do you determine whether the extra parameters are uid or scene id?

memetrollsXD: The order of the parameters: I recommend letting xyz go first, then scene id and then UID. You can check getPlayerByUID

omg-xtao: But scene id should be optional, just like UID.

omg-xtao: Using player id should not turn the scene id into a must parameter.

memetrollsXD:
That’s also how it goes with the /give command. Some arguments are optional
but you must provide the other argument even if they’re all optional

omg-xtao:
> That’s also how it goes with the /give command. Some arguments are optional but you must provide the other argument even if they’re all optional

I looked at the way the give command is implemented and the code does not work when the uid and the item id are duplicated.

For example, `give 223 1`, if the uid `233` exists, it will give me an item `223` , but what I actually want is to give `233` an item 1

memetrollsXD (approved):
I guess there is no other way

ghost:
Please fix your conflicts so we can merge.

omg-xtao:
> Please fix your conflicts so we can merge.

Done.

PR #354, opened 2022-04-29
Add 'refinement' option to 'give' command

RealHeart:
Example: !give 15502 1 90 5


memetrollsXD:
Wouldn´t this clash with `/give <weaponID> [amount] [level]`? There is no refinement option, but that can easily be added

RealHeart:
> Wouldn´t this clash with `/give <weaponID> [amount] [level]`? There is no refinement option, but that can easily be added

Add 'refinement' option to 'give' command

memetrollsXD (approved):

PR #362, opened 2022-04-29
Give Artifact: Clear random props first before adding all picked props

exzork:
I run command giveart with following stat
```
/giveart 77524 50990 965002 965004,10 965007,10 965009 21
```
It will give me this artifact
image
however this is not what command above says because propsIdList is appended without clearing it first. so this PR will add that.
After removing propsIdList first this is what i get 
image


ghost (approved):

Secack:
It never happened to me😳Sorry for this

exzork:
> It never happened to me😳Sorry for this

np, quick fix. Maybe in later commit there were change in how itemData works

PR #372, opened 2022-04-30
Add Language Resource Properties Read Config LANGUAGE_LOCAT Attr TO L…

TangHuLuTaiTian:
…oad Properties File


Scirese:
Why not add a function to let translators can use variables when they writing properties file?
You know,sometime we need to adjust some word order when we are translating these strings,and we cannot just only translate these fixed and common words...

PR #373, opened 2022-04-30
Monsters should drop items in world

MrAru:
fix the typo introduced in https://github.com/Grasscutters/Grasscutter/pull/366/commits/f1934c3a47d6c7ef888581bcf7f361d063154215


memetrollsXD (approved):

rushiawa:
they did not drop again in dungeon?

Birdulon:
> they did not drop again in dungeon?

They were dropping in domains and it was pretty great. Intended behaviour is that they don't drop in domains which is what this PR fixed.

PR #381, opened 2022-04-30
List command should show player uid

Coooookies:
/list command show player uid


KingRainbow44 (changes requested):
Can you add an option for listing player UIDs with nicknames?

Coooookies:
> Can you add an option for listing player UIDs with nicknames?

It means "/list" or "/list uid" have different result?
DDEZHH TO9%QXCD}4X8%{ 4

Coooookies:
> Can you add an option for listing player UIDs with nicknames?

It means "/list" or "/list uid" have different result?
image

KingRainbow44 (approved):

PR #383, opened 2022-04-30
Complete rework of Dispatch, Added DebugModes

4Benj:
\* I'm purposefully pushing to API branch because these changes were made for plugin developers
Changes:
- Reworked dispatch completely
      Now uses [java-express](https://github.com/Aarkan1/java-express) instead of HTTPServer. 
      This allows for plugin developers to more easily add routes/endpoints to the dispatch server and just makes the dispatch server more cleaner in general.
      Note: java-express is just [javalin](https://javalin.io/documentation) made to look like ExpressJS. In my opinion express is quite easily to learn compared to some other solutions, so having something similar to it for Java might be very helpful for others.
- Added Debug Modes to the config
      This replaces `LOG_PACKETS` and allows people to switch from logging no packets and requests, to only missing ones, and all of them. [ALL, MISSING, NONE]
      I've wanted to do this for a while but just wasn't given enough of a reason
- exclude `plugins` folder
- `*.jar` files are still excluded except the ones inside the `lib` folder (I needed to add java-express to the `lib` folder)


KingRainbow44:
totally not like i didnt see this so that's y I'm merging it into dev

KingRainbow44 (approved):

memetrollsXD (approved):

mingjun97:
This commit makes my client crash at the login stage.

I tried to download artifacts from the development github action build. The one before this commit works fine but this one got crash.

Mien client was iOS/iPadOS client on M1 MAC using Playcover.

image

KingRainbow44:
word of advice, say what branch your on (it isn't obvious by the image) and don't use builds from 7 hours ago

mingjun97:
Sorry. I should made it clear. 

I'm on development branch. 

The reason why I was trying to use builds from 7 hours ago is that the latest build from development branch crashes as well and I did a simple binary search to figure out which commit trigger the issue.

KingRainbow44:
the latest build *should* work (i just tested it)
go [here](https://discord.com/channels/965284035985305680/967250975997980682) for further support on this matter pls

PR #392, opened 2022-05-01
Fix crash when login

mingjun97:
Dispatch server route `POST /hk4e_global/mdk/shield/api/verify` may return `"email": null,` in its response, which will trigger crash in some client.

Add a guard to replace `null` with empty string will prevent such kind of crash.

### DIFF
--- a/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
+++ b/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
@@ -381,6 +381,9 @@ public final class DispatchServer {
 				responseData.data.account.uid = requestData.uid;
 				responseData.data.account.token = requestData.token;
 				responseData.data.account.email = account.getEmail();
+				if (responseData.data.account.email == null) { // null will trigger crash in some client
+					responseData.data.account.email = "";
+				}
 
 				Grasscutter.getLogger().info(String.format("[Dispatch] Client %s logged in via token as %s",
 						req.ip(), responseData.data.account.uid));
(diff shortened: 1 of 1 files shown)


mingjun97:
Wait.. Forget to revert other useless changes for debugging purpose.

mingjun97:
Revert my changes for debugging purpose. Should be fine now.

The `/hk4e_global/combo/granter/api/compareProtocolVersion` should be a GET request form the packet I captured. Thus I marked it with `get` method.

The `/admin/mi18n/plat_oversea/m202003048/m202003048-version.json` is a endpoint that I got log saying missing thus I just add it.

ghost (approved):

4Benj:
> The `/hk4e_global/combo/granter/api/compareProtocolVersion` should be a GET request form the packet I captured. Thus I marked it with `get` method.

Are you sure? on my end compareProtocolVersion is a POST request

4Benj:
> Are you sure? on my end compareProtocolVersion is a POST request
image

4Benj:
> Dispatch server route `POST /hk4e_global/mdk/shield/api/verify` may return `"email": null,` in its response, which will trigger crash in some client.

This is weird because apart from very minor changes, I never touched the verify route

mingjun97:
> > Are you sure? on my end compareProtocolVersion is a POST request
> 
> image

image

Then I guess it could be either GET or POST.. Let's make it as `any` then. Let me open another PR to fix it. Additional, there exsit another crash point. I'll fix them together.




> > Dispatch server route `POST /hk4e_global/mdk/shield/api/verify` may return `"email": null,` in its response, which will trigger crash in some client.
> 
> This is weird because apart from very minor changes, I never touched the verify route

The response(version that not trigger crash) before your commit omit the `email` field actually. Maybe the httpServer on top of it omitted the `null` field? TBH, I don't know too much about Java.

4Benj:
> Then I guess it could be either GET or POST.. Let's make it as any then. Let me open another PR to fix it.

Yeah that makes sense, honestly to me a POST makes the most sense to compare protocol versions, maybe GET is for comparing it on the client instead of the server.

> The response(version that not trigger crash) before your commit omit the email field actually. Maybe the httpServer on top of it omitted the null field? TBH, I don't know too much about Java.

Yeah, I think you are right and that is probably is why it is crashing

PR #393, opened 2022-05-01
implement McoinExchange packet Handler

CCasusensa:
preview:
https://i.imgur.com/fvPGoJr.mp4


ghost (approved):
Good work, but as a future note, a GameSession object will never be null in a packet handler, so no need to put that null check there.

PR #394, opened 2022-05-01
Fix crash and revise route

mingjun97:
* Fix another point which will trigger iOS client to crash when login
* Revise `compareProtocolVersion` route to handle all method

PR #396, opened 2022-05-01
Cleaned up dispatch iOS fixes

4Benj:
While #392 and #394 fixed the issues iOS devices were having with logging in, it left if statements everywhere and if there comes a time when another login route needs to get created in the future this issue may arise again.
This PR clean up the if statements by moving them into the `getEmail()` which now if a player's email is not set or is empty, `getEmail()` should always return "".

Thanks @mingjun97 for helping to test

### DIFF
--- a/src/main/java/emu/grasscutter/game/Account.java
+++ b/src/main/java/emu/grasscutter/game/Account.java
@@ -74,7 +74,11 @@ public class Account {
 	}
 	
 	public String getEmail() {
-		return email;
+		if(email != null && !email.isEmpty()) {
+			return email;
+		} else {
+			return "";
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


src/main/java/emu/grasscutter/game/Account.java
```diff
@@ -74,7 +74,11 @@ public void setPlayerId(int playerId) {
 	}
 	
 	public String getEmail() {
-		return email;
+		if(email != null && !email.isEmpty()) {
```
ghost, line 77: Just a side note: email.isEmpty() is the same as the empty string you are returning, so you dont need it in this if statement.

ghost (approved):

PR #407, opened 2022-05-01
Add `/setstats mhp` to set Max HP

jie65535:
The Max HP should be modified.


memetrollsXD (changes requested):
Please add an extra case for max HP instead of replacing the cur HP one

Lynricsy:
> Please add an extra case for max HP instead of replacing the cur HP one

He has already fixed and please merge this PR.I've expected for this feature for a long time.Thanks.

ghost (approved):

memetrollsXD (approved):

PR #408, opened 2022-05-01
Fixed can set talent level to 16 bug

jie65535:
Level should be **lower than 16**

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/TalentCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/TalentCommand.java
@@ -53,7 +53,7 @@ public final class TalentCommand implements CommandHandler {
                             CommandHandler.sendMessage(sender, "To get talent ID: /talent getid");
                             return;
                         }
-                        if (nextLevel > 16){ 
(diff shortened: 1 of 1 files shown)


alt3ri:
haven't it corrected already? the way you add the = just make the character talent getting locked

memetrollsXD:
Talent level should be 15 or lower due to client-side restrictions

jie65535:
The problem now is that the talent level can be set to 16, my commit fixed that.
Please review again

memetrollsXD (approved):

Birdulon:
What happens clientside for a level of 0, or negative numbers?

memetrollsXD:
It also locks the talents I’d imagine

alt3ri:
oops, sorry. Talent lv15 only available in event

memetrollsXD:
Which characters? Max is 15..

jie65535:
# Set talent level to 15
Set talent level to 15

# Set talent level to 16
Set talent level to 16

alt3ri:
that's the reason. I did correct it already. Using >= just make it doesn't right at all

PR #413, opened 2022-05-01
Implement kick command (TEST ONLY)

Asnxthaony:
It's just a temporary solution, needs to be improved later.


KingRainbow44:
can you move the disconnect packet to the player or some other class by chance? instead of just adding it specifically in the kick command

ghost:
> can you move the disconnect packet to the player or some other class by chance? instead of just adding it specifically in the kick command

It should be fine for now, we gotta update our kcp library anyways, its missing some things.

ghost:
Kcp disconnect is now supported (albiet without an option to send close code).

PR #417, opened 2022-05-01
Implement gacha history record subsystem

mingjun97:
* Frontend is not very beautiful yet
* Didn't include too much `some anime game` data in the page to avoid being DMCA'd

The frontend could be refined without modifications towards the jar server.

image
image

image

image

### DIFF
--- a/data/gacha_records.html
+++ b/data/gacha_records.html
@@ -1,0 +1,176 @@
+<html>
+    <head>
+        <script>
+            // Debug entry
+            // record = [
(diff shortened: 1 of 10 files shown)


ghost (commented):
The html template should be in the [src/resources](https://github.com/Grasscutters/Grasscutter/tree/development/src/main/resources) folder in my opinion, other than that, good work.

lunaticwhat:
nice one I was thinking about this and now atleast we have a cool backend ready !

ghost (approved):

PR #422, opened 2022-05-02
Refactor SetStats command

Birdulon:
Attempts to make it simpler to extend with future stats, and includes some additional ones.

Additionally adds support to target different players via a `@UID` argument which is optional in-game but required from the server console.

Specifically, replaces a lengthy `switch` statement with large amounts of code duplication, with a `HashMap` definition of all possible stats. All `FightProperty` stats are included for developers, though only a subset are documented with user-friendly names on them, a few more than before these changes.

TODO:
 1. ~I'm somewhat unfamiliar with idiomatic Java so I'm not sure where exactly I should put the definitions map. I construct it in the `execute()` function for now since it proves the concept, but it should be created once and stored, not created every time the command is run.~ Fixed.
 2. I copied the @UID idea from #353 but went with `@UID` instead of `@player id` which is a consistency break. The docs on #353 use `@playerUid` in the readme, and I think all of these should be the same.
- - @UID arguments would be nice to have on all commands which currently can't be used on the server console, I'm willing to go over those next.
- - Perhaps it would make sense to add more permissions such that targeting other players in-game requires a higher permission than just using it on themselves?
3. ~How should localisation be handled? Probably ties in with 1. Maybe mapping files for the keys of form `SHIELD_COST_MINUS_RATIO`: "Shield Strength" for each language?~

### DIFF
--- a/src/main/java/emu/grasscutter/Language.java
+++ b/src/main/java/emu/grasscutter/Language.java
@@ -204,41 +204,51 @@ public final class Language {
     public String SetFetterLevel_invalid_fetter_level = "Invalid fetter level.";
 
     // SetStats
-    public String SetStats_usage = "Usage: setstats|stats <stat> <value>";
-    public String SetStats_setstats_help_message = "Usage: /setstats|stats <hp | mhp | def | atk | em | er | crate | cdmg> <value> for basic stats";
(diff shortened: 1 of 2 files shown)


Birdulon:
Updated.

Birdulon:
Updated against 9407e63

ghost (approved):

PR #427, opened 2022-05-02
Fix announcement display

BaiSugar:
Fix announcement display, if there is a better way to add announcement please let me know
WRN@XXY VS8`36755TH9J E


BaiSugar:
If you can't see the picture, please try to no block https://uploadstatic-sea.mihoyo.com

ghost (approved):

PR #430, opened 2022-05-02
add defaultPermissions

easy-do:
Currently, the creating user does not store the default permissions in the database, resulting in the configuration item having no effect


ghost:
Thank you for making your pull request. However, default permissions is for the `AutomaticallyCreateAccounts` option in the config, [which works as intended](https://github.com/Grasscutters/Grasscutter/blob/d912b59d93b09ac67bb8c70dd6301bd9da688784/src/main/java/emu/grasscutter/server/dispatch/authentication/DefaultAuthenticationHandler.java#L44). Console created accounts will and should always have all permissions by default because the person creating the account already has access to the console.

PR #158, opened 2022-04-23
authentication system

exzork:
login : dispatchserver/grasscutter/login
register: dispatchserver/grasscutter/register

You get JWT token, payload are token, username, and uid.
and copy token to username field when logging in

PR #444, opened 2022-05-02
Fix the existing users can't create password

omg-xtao:
Users who had already set up before GCAuth was implemented would be unable to do register.

The commit resolves this issue.

### DIFF
--- a/src/main/java/emu/grasscutter/database/DatabaseHelper.java
+++ b/src/main/java/emu/grasscutter/database/DatabaseHelper.java
@@ -57,7 +57,13 @@ public final class DatabaseHelper {
 		// Unique names only
 		Account exists = DatabaseHelper.getAccountByName(username);
 		if (exists != null) {
-			return null;
+			if (exists.getPassword() == null) {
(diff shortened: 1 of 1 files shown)


ghost:
I mainly left that function as a filler for when a vanilla auth system could be implemented. Also passwords should never be stored in plaintext.

omg-xtao:
> I mainly left that function as a filler for when a vanilla auth system could be implemented. Also passwords should never be stored in plaintext.

Maybe there is a problem with my description.

Users who were created using the createAccountWithId function will not be able to use the createAccountWithPassword method to create a password.

This is independent of the password's storage method.

4Benj:
Unless something changed between #158 and GCAuth, you should just be able to use the change password route to set a password. However, if you cannot, this is most likely an issue with GCAuth, not Grasscutter.

omg-xtao:
> Unless something changed between #158 and GCAuth, you should just be able to use the change password route to set a password. However, if you cannot, this is most likely an issue with GCAuth, not Grasscutter.

ok.

PR #455, opened 2022-05-03
Implement food heal function

omg-xtao:


hit2004:
but when the heal abilities turns normal?

ghost (approved):

PR #472, opened 2022-05-03
Optimize multilingualism (using variables)

577fkj:
Optimize multilingualism (using variables)


ghost:
You have merge conflicts because youre trying to PR an old commit that we already pushed, can you make a new PR with only your changes? Thank you.

PR #480, opened 2022-05-03
Add Chinese Translation

Scirese:
This pr relies on #472


Scirese:
Some place are not translated because i really don't now how to translate them:(, maybe some one will fix them

ghost:
I already accepted another PR for this, thank you for your effort tho.

PR #492, opened 2022-05-04
Optimize multilingualism

577fkj:


src/main/java/emu/grasscutter/command/commands/TeleportCommand.java
```diff
@@ -82,7 +82,7 @@ public void execute(Player sender, List<String> args) {
             if (!result) {
                 CommandHandler.sendMessage(sender, Grasscutter.getLanguage().Teleport_invalid_position);
             } else {
-                CommandHandler.sendMessage(sender, String.format(Grasscutter.getLanguage().Teleport_message, targetPlayer.getNickname(), x, y, z, sceneId));
+                CommandHandler.sendMessage(sender, Grasscutter.getLanguage().Teleport_message.replace("{name}", targetPlayer.getNickname()).replace("{x}", Float.toString(x)).replace("{y}", Float.toString(y)).replace("{z}", Float.toString(z)).replace("{id}", Integer.toString(sceneId)));
```
Birdulon, line 85: Is there a point to replacing everything with keyword replaces? It seems particularly painful in Java.

577fkj: Support word order

Scirese: Some times we need to adjust some word order,and also need to adjust the order of some vars.
If not do that,the vars' order are fixed and unadjustable. It seems particularly painful for translators too.

Birdulon: Ok, seems like this is the cleanest way in Java.

ghost (approved):

PR #499, opened 2022-05-04
[Breaking] Command Targeting overhaul

Birdulon:
Thanks @Melledy for the previous merge :)

This is an incomplete rework of the commands system to introduce the concept of a `targetPlayer` for all commands to make use of.
If the syntax is acceptable, I'll go ahead and fix all the Commands to use it.

- Any argument starting with `@` will be treated as a target player UID, for any command. This argument will be deleted from the list and commands won't see it. This has the highest priority.
- If messaging in-game, you can direct message commands to other players with the usual prefixes and they will be the `targetPlayer`.
- Using the command `/target [UID]` will set the default `targetPlayer` for all future commands to `UID`, or delete this default if none is given.
- If none of the above give a `targetPlayer`, the sender will be used, or `null` for commands entered in the console. It is up to individual commands to complain about no `targetPlayer` being present if they cannot be run from the console without a targeted player.
- Players targeting other players in-game may require additional permissions on a per-command basis. This does not apply to the console.

Examples:
`stats @10001 hp 100`: under all circumstances, this targets player with UID `10001`.

Set a target player and do multiple commands on them from the console:
```
target 10001
stats maxhp 500
stats hp 500
stats respyro 35%
```

Do the same in-game and then something to yourself:
```
!target 10002
!stats maxhp 500
!stats hp 500
!stats respyro 35%
!target
!giveall
```

Also works with `@[UID]` instead of `target [UID]`
```
@10001
killall
teleport 0 0 0
@
```

### DIFF
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -213,7 +213,7 @@ public final class Grasscutter {
 
 			isLastInterrupted = false;
 			try {
-				CommandMap.getInstance().invoke(null, input);
+				CommandMap.getInstance().invoke(null, null, input);
 			} catch (Exception e) {
 				Grasscutter.getLogger().error(language.Command_error, e);
 			}
--- a/src/main/java/emu/grasscutter/command/Command.java
+++ b/src/main/java/emu/grasscutter/command/Command.java
@@ -14,6 +14,8 @@ public @interface Command {
     String[] aliases() default {};
 
     String permission() default "";
+    
+    String permissionTargeted() default "";
 
     boolean threading() default false;
 }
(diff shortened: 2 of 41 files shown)


ghost:
This is a good way of isolating the player uid from commands.

Birdulon:
Ready for testing, going to sleep for now though since there's so many commands to test.
Might have refactored some commands a bit too aggressively for the scope of the PR, some definitely needed it though.

ZYFDroid:
I think a command like `runas <Player> <Command>` might be a better idea, less modification to current code

Birdulon:
> I think a command like `runas <Player> <Command>` might be a better idea, less modification to current code

"Why not add another layer of hacks?"

Existing plugin argument parsing: 
image

Rewrite:
image

The current state of commands that target users is not worth working around forever.

Birdulon:
I've tested every command against simple arguments (including `/account delete`, `/coop` and `/tpall`, and testing previously-ingame-only commands from the console) but there might be some corner cases on more specific argument sets for some commands.
Some language stuff will need updating afterwards too, but that seems to be changing a lot currently so I held off on touching the CN file.

ghost (approved):

PR #519, opened 2022-05-04
Implement map marking features

gentlespoon:
Teleport still exists on fish hook mark.
Added mapMark-related protos.
Map marking data is stored in players collection.

### DIFF
--- a/proto/MapMarkTipsInfo.proto
+++ b/proto/MapMarkTipsInfo.proto
@@ -1,0 +1,10 @@
+syntax = "proto3";
+
+option java_package = "emu.grasscutter.net.proto";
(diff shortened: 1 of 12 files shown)


ZH950219:
1

ghost (approved):

ghost:
Good work, but as a sidenote, it would be better to put map marks in a different mongo table, so if the server had to load offline players, it wouldnt need to load all the marks unless called for. Its not a big deal however.

gentlespoon:
Agreed.
Perhaps we can make a new collection for map marks and other future data that are player-related but are less frequently used.

PR #575, opened 2022-05-06
Target permissions

Birdulon:
image

Adds permissionTargeted to all applicable commands. (i.e. every command which targets a player and isn't sendmail)

The `player.perm` and `server.perm` distinction seems a bit ill-defined, things like `drop` seem like they should be `player.drop` instead of `server.drop` (ideally I think the true `server.perm` commands wouldn't have a `target.perm` permission at all since they would imply usage on anyone and everyone), but that might be something to leave for one big permissions rework down the track since any changes would involve db migration procedures for server operators. I think some wildcard matching would be good at some point, so admins can set things like `player.*` and `*.spawn` for convenience.

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/ChangeSceneCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/ChangeSceneCommand.java
@@ -8,7 +8,7 @@ import java.util.List;
 
 import static emu.grasscutter.utils.Language.translate;
 
-@Command(label = "changescene", usage = "changescene <scene id>", aliases = {"scene"}, permission = "player.changescene", description = "commands.changescene.description")
+@Command(label = "changescene", usage = "changescene <scene id>", aliases = {"scene"}, permission = "player.changescene", permissionTargeted = "player.changescene.others", description = "commands.changescene.description")
 public final class ChangeSceneCommand implements CommandHandler {
 
     @Override
--- a/src/main/java/emu/grasscutter/command/commands/ClearCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/ClearCommand.java
@@ -14,7 +14,7 @@ import static emu.grasscutter.utils.Language.translate;
 
 @Command(label = "clear", usage = "clear <all|wp|art|mat>", //Merged /clearartifacts and /clearweapons to /clear <args> [uid]
         description = "commands.clear.description",
-        aliases = {"clear"}, permission = "player.clearinv")
+        aliases = {"clear"}, permission = "player.clearinv", permissionTargeted = "player.clearinv.others")
 
 public final class ClearCommand implements CommandHandler {
 
--- a/src/main/java/emu/grasscutter/command/commands/CoopCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/CoopCommand.java
@@ -9,7 +9,7 @@ import java.util.List;
 
 import static emu.grasscutter.utils.Language.translate;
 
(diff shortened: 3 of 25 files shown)


jie65535:
My suggestion is to use `@` as default permission

```java
String permissionTargeted() default "@";
```

memetrollsXD:
How about a new subpermission? `player.clear.others` for all commands

memetrollsXD:
It's basically just another syntax of what you have here, looking for peoples opinions

lhhxxxxx:
> My suggestion is to use `@` as default permission
> 
> ```java
> String permissionTargeted() default "@";
> ```

It's a good idea. Only those who have the permission "@" can use @ in command.

lhhxxxxx:
Account.hasPermission() should also be updated.

Birdulon:
> Account.hasPermission() should also be updated.

Good catch, I didn't see that it already has wildcards

lhhxxxxx:
> > Account.hasPermission() should also be updated.
> 
> Good catch, I didn't see that it already has wildcards

I don't consider subpermission to be a good idea, because "player.*" seems to cover all the subpermissions under the player permission, but in fact we don't want those with player permission to have @ permission

lhhxxxxx:
```
public boolean hasPermission(String permission) {
		return this.permissions.contains(permission) ||
                this.permissions.contains("*") ||
                (this.permissions.contains("player") || this.permissions.contains("player.*")) && permission.startsWith("player.") ||
                (this.permissions.contains("server") || this.permissions.contains("server.*")) && permission.startsWith("server.") ||
                (this.permissions.contains("target") || this.permissions.contains("target.*") || this.permissions.contains("@")) && permission.startsWith("target.");
	}
```
This could be a solution

4Benj:
The permission system will be getting a rewrite soon to make it more expandable and to fix this sort of thing. Current plan is for * to be a first layer wildcard and ** to be an all layer wildcard

Birdulon:
Added this
> Current plan is for * to be a first layer wildcard and ** to be an all layer wildcard

EDIT: Hacky unit test over at https://github.com/Birdulon/Grasscutter/tree/TargetPermsUnitTest
image

ghost:
> The permission system will be getting a rewrite soon to make it more expandable and to fix this sort of thing. Current plan is for * to be a first layer wildcard and ** to be an all layer wildcard

Note that `player.*.self` and `player.*.target` might be a good alternative and will be simpler to implement than what you propose. Examples include:
- `player.*` - anything that starts with `player.`
- `player.*.target` - anything that matches `player.*.target`, where `*` is a wildcard
- `player.*.target.*` - anything that starts with `player.*.target.`, where `*` is a wildcard
- `player.coop.self.*` - anything that starts with `player.coop.self.`, where `*` is a wildcard

Implementation of wildcard can be done as regular expression, so internally it becomes:
- `^player(?:\..*)?$`
- `^player\..*\.target$`
- `^player\..*\.target(?:\..*)?$`
- `^player\.coop\.self(?:\..*)?$`

Internally, we would store list of permissions account has, then escape expression and replace `\*` with `.*`, add `^` and `$` and then test it against full permission name in code. Note that last instance of `\.\*` should be replaced with `(?:\..*)?` instead so that it encompasses both `player.coop.target` as well as `player.coop.target.join`.

4Benj (approved):

ghost (approved):

Birdulon:
Fixed merge conflicts.

memetrollsXD (approved):

PR #580, opened 2022-05-06
Fix the problem that the reference of serverHook in Plugin object is null

Bwly999:
When PluginManager loading the plugins, the Serverhook has not been initialized, which will cause the ServerHook in the plugin object to be null


src/main/java/emu/grasscutter/server/game/GameServer.java
```diff
@@ -212,6 +202,23 @@ public void deregisterWorld(World world) {
 		
 	}
 
+	@Override
+	public synchronized void start() {
+		// Schedule game loop.
+		ScheduledExecutorService gameLoop = Executors.newScheduledThreadPool(2);
```
ghost: Why did you move the game loop and why is it in a thread pool..?

src/main/java/emu/grasscutter/Grasscutter.java
```diff
@@ -92,14 +92,13 @@ public static void main(String[] args) throws Exception {
...
-		// Create plugin manager instance.
-		pluginManager = new PluginManager();
-
 		// Create server instances.
 		dispatchServer = new DispatchServer();
 		gameServer = new GameServer(new InetSocketAddress(getConfig().getGameServerOptions().Ip, getConfig().getGameServerOptions().Port));
 		// Create a server hook instance with both servers.
 		new ServerHook(gameServer, dispatchServer);
+		// Create plugin manager instance.
+		pluginManager = new PluginManager();
```
4Benj, line 105: I think there might have been a reason for plugins to be initialized before the dispatch or game server.
I'm guessing it's to prevent people from editing the server's functions from ``onLoad`` and to force people to do it in ``onEnable`` but I'm not exactly sure.
@KingRainbow44 can tell you if this was intentional or not and if it is safe to move.

src/main/java/emu/grasscutter/server/game/GameServer.java
Bwly999: Because onTick function in game loop will call `Grasscutter.getPluginManager().invokeEvent(this);`,
but after moving the order of creating plugin manager, it will cause a null point exception because  it is not instantiated in this time。As for thread pool, it's the recommendation of IDEA....

src/main/java/emu/grasscutter/Grasscutter.java
Bwly999: Every plugin will be instantiated in plugin manager constructor, and in the time, if ServerHook.instance is still null, `private final ServerHook server = ServerHook.getInstance();` in Plugin  seems a little unless. Maybe need a new proper position to call this statement.

4Benj: Ohhhhh okay, I haven't had the chance of play with hooks yet, mb.
I just thought you were trying to call getDispatchServer or getGameServer from onLoad or something.
ik the PR is about the serverHook but I didn't know it worked like that

src/main/java/emu/grasscutter/server/game/GameServer.java
ghost: Makes sense but theres no need to put the gameloop in a thread pool.

Bwly999: okay, I've rolled it back

KingRainbow44 (approved):

PR #582, opened 2022-05-06
Stop WindSeedClientNotify and PlayerLuaShellNotify from being sent

4Benj:
Since ``WindSeedClientNotify`` allows remote code execution, I think our best bet is to stop it from being sent to clients at all.
I know this is very easy to remove but I don't think there is much else we can currently do unless we find a way of validating code before sending to client...


memetrollsXD (changes requested):
Please add PlayerLuaShellNotify to this list as well

4Benj:
> Please add PlayerLuaShellNotify to this list as well

Done

memetrollsXD (approved):

PR #585, opened 2022-05-06
修复一个Gson发序列化json文件的空值问题

cfuncode:


ClayCoffee:
please use english

cfuncode:
> please use english

Fix up: A problem when Gson deserialization json file will be null value. 
because a lot of  properties of bean was Uppercase
and the json file key name was not Uppercase

PR #590, opened 2022-05-06
Implement Avatar Expedition System

CCasusensa:


ghost (approved):
I like that you guys seperated the packet handler functions into a manager, which i usually do to make things cleaner. Thank you very much for this commit.

Issue #581, opened 2022-05-06
Spiral Abyss start error


MapleNe:
java.lang.NullPointerException: Cannot invoke "emu.grasscutter.game.player.Player.getTeamManager()" because "this.player" is null
        at emu.grasscutter.game.tower.TowerManager.teamSelect(TowerManager.java:28)
        at emu.grasscutter.server.packet.recv.HandlerTowerTeamSelectReq.handle(HandlerTowerTeamSelectReq.java:22)
        at emu.grasscutter.server.game.GameServerPacketHandler.handle(GameServerPacketHandler.java:85)
        at emu.grasscutter.server.game.GameSession.onMessage(GameSession.java:242)
        at emu.grasscutter.netty.KcpChannel.channelRead(KcpChannel.java:43)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
        at io.jpower.kcp.netty.UkcpServerChannel$UkcpServerUnsafe.read(UkcpServerChannel.java:603)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:722)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:658)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:584)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:496)
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
[...]

PR #594, opened 2022-05-06
Tower Dungeons Handoff between levels

Akka0:
- User can continue to enter the next level dungeon in towers
- fix the NPE in issue #581 
- fix the level dungeon confusion
- Now users will be forced to leave dungeon when they log out


ghost (approved):
Good work!

PR #619, opened 2022-05-07
Fix failed to load language file:  zh-CN.json

xfl03:
## Exception related
```
com.google.gson.stream.MalformedJsonException: Invalid escape sequence at line 1 column 8245 path $.commands.setStats.help_message
```


xfl03:
#618

PR #621, opened 2022-05-07
Fix incorrect language tag in init

xfl03:
When use `Locale.getDefault().toLanguageTag()`, it will be `zh-Hans-CN` for Chinese. But `zh-Hans-CN.json` cannot be found, using `zh-CN` is correct.

### DIFF
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -149,14 +149,8 @@ public final class Grasscutter {
 
 	public static void loadLanguage() {
 		var locale = config.LocaleLanguage;
-		String languageTag = locale.toLanguageTag();
-		if (languageTag.equals("und")) {
-			Grasscutter.getLogger().error("Illegal locale language, using en-US instead.");
-			language = Language.getLanguage("en-US");
-		} else {
-			languageTag = String.format("%s-%s", locale.getLanguage(), locale.getCountry());
-			language = Language.getLanguage(languageTag);
-		}
+		String languageTag = String.format("%s-%s", locale.getLanguage(), locale.getCountry());
+		language = Language.getLanguage(languageTag);
 	}
 
 	public static void saveConfig() {
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -154,7 +154,7 @@ public final class Grasscutter {
 			Grasscutter.getLogger().error("Illegal locale language, using en-US instead.");
 			language = Language.getLanguage("en-US");
 		} else {
-			languageTag = String.format("%s-%s", locale.getCountry(), locale.getLanguage());
+			languageTag = String.format("%s-%s", locale.getLanguage(), locale.getCountry());
 			language = Language.getLanguage(languageTag);
 		}
 	}
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -154,6 +154,7 @@ public final class Grasscutter {
 			Grasscutter.getLogger().error("Illegal locale language, using en-US instead.");
 			language = Language.getLanguage("en-US");
 		} else {
+			languageTag = String.format("%s-%s", locale.getCountry(), locale.getLanguage());
 			language = Language.getLanguage(languageTag);
 		}
 	}


src/main/java/emu/grasscutter/Grasscutter.java
```diff
@@ -154,6 +154,7 @@ public static void loadLanguage() {
 			Grasscutter.getLogger().error("Illegal locale language, using en-US instead.");
 			language = Language.getLanguage("en-US");
 		} else {
+			languageTag = String.format("%s-%s", locale.getCountry(), locale.getLanguage());
```
jie65535: `country`-`language` ?
`CN`-`zh` ?

xfl03: Thanks for your review, this bug has been fixed.

jie65535:
I think it is better to follow the IETF BCP 47 standard.

Do not change the [default behavior](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html).

xfl03:
> default behavior

The reason why I modified it can be seen as follows.
## Run the server first time 
It will call `Locale.getDefault().toLanguageTag()` and get `zh-Hans-CN`. But `zh-Hans-CN.json` cannot be found.  
Moreover, the `Locale.getDefault()` will be written to `config.json` as `zh_CN_#Hans`.   
For more details, `Locale.getDefault()` will return a `Locale` object with a `BaseLocale(language = "zh", script = "Hans", region = "CN")`.
## Next run
`zh_CN_#Hans` will be read and parse into `Locale` in `Config.localeLanguage`, but the `Config.localeLanguage.toLanguageTag()` will get `zh-CN` instead of `zh-Hans-CN`.  
For more details, `Config.localeLanguage` will has a `Locale` object with a `BaseLocale(language = "zh", script = "", region = "CN", variant = "#Hans")`.

As the result, I think make it to `zh-CN` is a better choice.

iTruth:
since #612, getLanguage will return default language if it can't find locale language.  
so there is no necessary to add these code now

xfl03:
> #612

This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.

iTruth:
> > #612
> 
> This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.

ru sure about that? have u ever tried?

xfl03:
> > > #612
> > 
> > 
> > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> 
> ru sure about that? have u ever tried?

Yes, I am sure.  
Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743 .
If you want to reproduce this, just remove `config.json` and start the server.

iTruth:
> > > > #612
> > > 
> > > 
> > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > 
> > ru sure about that? have u ever tried?
> 
> Yes, I am sure.  
> Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743 .
> If you want to reproduce this, just remove `config.json` and start the server.

still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution

xfl03:
> 

`Hant` will be `zh-TW` or `zh-HK`, isn't influenced.
The OS I used is macOS 12.3.1. Could you reproduce in macOS?

xfl03:
> `Hant` will be `zh-TW` or `zh-HK`, isn't influenced.
> The OS I used is macOS 12.3.1. Could you reproduce in macOS?

`Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
The OS I used is macOS 12.3.1. Could you reproduce in macOS?

xfl03:
> > > > > #612
> > > > 
> > > > 
> > > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > > 
> > > 
> > > ru sure about that? have u ever tried?
> > 
> > 
> > Yes, I am sure.
> > Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in [#621 (comment)](https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743) .
> > If you want to reproduce this, just remove `config.json` and start the server.
> 
> still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution

`Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
The OS I used is macOS 12.3.1. Could you reproduce in macOS?

The debug screenshot in macOS can be seen following.
unknown

iTruth:
> > > > > > #612
> > > > > 
> > > > > 
> > > > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > > > 
> > > > 
> > > > ru sure about that? have u ever tried?
> > > 
> > > 
> > > Yes, I am sure.
> > > Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in [#621 (comment)](https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743) .
> > > If you want to reproduce this, just remove `config.json` and start the server.
> > 
> > still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution
> 
> `Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
> The OS I used is macOS 12.3.1. Could you reproduce in macOS?
> 
> The debug screenshot in macOS can be seen following.
> unknown
> 
> 

ok, I know there's an issue here. at the beginning what I want to say is pr611 is useless since pr612. so there's no necessary for u to fix this at pr611. maybe fix in Language.java is better?
for now, u said Hant isn't influenced, I want to know why? can u guarantee that Hant will never appear?

xfl03:
> > > > > > > #612
> > > > > > 
> > > > > > 
> > > > > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > > > > 
> > > > > 
> > > > > ru sure about that? have u ever tried?
> > > > 
> > > > 
> > > > Yes, I am sure.
> > > > Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in [#621 (comment)](https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743) .
> > > > If you want to reproduce this, just remove `config.json` and start the server.
> > > 
> > > 
> > > still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution
> > 
> > 
> > `Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
> > The OS I used is macOS 12.3.1. Could you reproduce in macOS?
> > The debug screenshot in macOS can be seen following.
> > unknown
> 
> ok, I know there's an issue here. at the beginning what I want to say is pr611 is useless since pr612. so there's no necessary for u to fix this at pr611. maybe fix in Language.java is better? for now, u said Hant isn't influenced, I want to know why? can u guarantee that Hant will never appear?

According to the file name in `resources/languages` ([link](https://github.com/Grasscutters/Grasscutter/tree/development/src/main/resources/languages)), there are `en-US`, `zh-CN` and `zh-TW`, not `en-Latn-US`, `zh-Hans-CN` or `zh-Hant-TW`.
Make language tag format to `Language-Country` can simply match the file name.

iTruth:
> > > > > > > > #612
> > > > > > > 
> > > > > > > 
> > > > > > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > > > > > 
> > > > > > 
> > > > > > ru sure about that? have u ever tried?
> > > > > 
> > > > > 
> > > > > Yes, I am sure.
> > > > > Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in [#621 (comment)](https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743) .
> > > > > If you want to reproduce this, just remove `config.json` and start the server.
> > > > 
> > > > 
> > > > still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution
> > > 
> > > 
> > > `Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
> > > The OS I used is macOS 12.3.1. Could you reproduce in macOS?
> > > The debug screenshot in macOS can be seen following.
> > > unknown
> > 
> > ok, I know there's an issue here. at the beginning what I want to say is pr611 is useless since pr612. so there's no necessary for u to fix this at pr611. maybe fix in Language.java is better? for now, u said Hant isn't influenced, I want to know why? can u guarantee that Hant will never appear?
> 
> According to the file name in `resources/languages` ([link](https://github.com/Grasscutters/Grasscutter/tree/development/src/main/resources/languages)), there are `en-US`, `zh-CN` and `zh-TW`, not `en-Latn-US`, `zh-Hans-CN` or `zh-Hant-TW`.
> Make language tag format to `Language-Country` can simply match the file name.

got it, before u fix this, can u revert pr611 first?

xfl03:
> > > > > > > > > #612
> > > > > > > > 
> > > > > > > > 
> > > > > > > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > > > > > > 
> > > > > > > 
> > > > > > > ru sure about that? have u ever tried?
> > > > > > 
> > > > > > 
> > > > > > Yes, I am sure.
> > > > > > Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in [#621 (comment)](https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743) .
> > > > > > If you want to reproduce this, just remove `config.json` and start the server.
> > > > > 
> > > > > 
> > > > > still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution
> > > > 
> > > > 
> > > > `Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
> > > > The OS I used is macOS 12.3.1. Could you reproduce in macOS?
> > > > The debug screenshot in macOS can be seen following.
> > > > unknown
> > > 
> > > 
> > > ok, I know there's an issue here. at the beginning what I want to say is pr611 is useless since pr612. so there's no necessary for u to fix this at pr611. maybe fix in Language.java is better? for now, u said Hant isn't influenced, I want to know why? can u guarantee that Hant will never appear?
> > 
> > 
> > According to the file name in `resources/languages` ([link](https://github.com/Grasscutters/Grasscutter/tree/development/src/main/resources/languages)), there are `en-US`, `zh-CN` and `zh-TW`, not `en-Latn-US`, `zh-Hans-CN` or `zh-Hant-TW`.
> > Make language tag format to `Language-Country` can simply match the file name.
> 
> got it, before u fix this, can u revert pr611 first?

Thanks for your understanding.
~~For instance, if there is a Japanese user which  locale is `ja-JP`, but `ja-JP.json` could not be found. At this time, the language file should be `en-US.json` as default, rather than no text would be displayed.~~
Sorry, that's pr 612's work.

iTruth:
> > > > > > > > > > #612
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > This bug will cause Chinese users get English translation when they run the server first time. It should be Chinese.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > ru sure about that? have u ever tried?
> > > > > > > 
> > > > > > > 
> > > > > > > Yes, I am sure.
> > > > > > > Chinese user will display Chinese only if run the server second time or later. But it will display English in the first run. The reason why cause this has been talked in [#621 (comment)](https://github.com/Grasscutters/Grasscutter/pull/621#issuecomment-1120163743) .
> > > > > > > If you want to reproduce this, just remove `config.json` and start the server.
> > > > > > 
> > > > > > 
> > > > > > still can't reproduce. everything works fine on my computer, but if u get Hans at first time I think u should also consider Hant. I don't think current way is the solution
> > > > > 
> > > > > 
> > > > > `Hant` will be `zh-TW` or `zh-HK`, which isn't influenced.
> > > > > The OS I used is macOS 12.3.1. Could you reproduce in macOS?
> > > > > The debug screenshot in macOS can be seen following.
> > > > > unknown
> > > > 
> > > > 
> > > > ok, I know there's an issue here. at the beginning what I want to say is pr611 is useless since pr612. so there's no necessary for u to fix this at pr611. maybe fix in Language.java is better? for now, u said Hant isn't influenced, I want to know why? can u guarantee that Hant will never appear?
> > > 
> > > 
> > > According to the file name in `resources/languages` ([link](https://github.com/Grasscutters/Grasscutter/tree/development/src/main/resources/languages)), there are `en-US`, `zh-CN` and `zh-TW`, not `en-Latn-US`, `zh-Hans-CN` or `zh-Hant-TW`.
> > > Make language tag format to `Language-Country` can simply match the file name.
> > 
> > got it, before u fix this, can u revert pr611 first?
> 
> Thanks for your understanding, but pr611 is necessary. 
> For instance, if there is a Japanese user which  locale is `ja-JP`, but `ja-JP.json` could not be found. At this time, the language file should be `en-US.json` as default, rather than no text would be displayed.

see pr612, that feature has already been implemented

lhhxxxxx:
You shouldn't delete the handle on "und".

xfl03:
> You shouldn't delete the handle on "und".

@iTruth think handle on `und` is unnecessary, so do I.
The fallback logic to `en-US` is in `Language.java`.

lhhxxxxx:
> > You shouldn't delete the handle on "und".
> 
> @iTruth think handle on `und` is unnecessary, so do I.
> The fallback logic to `en-US` is in `Language.java`.

und means the locale language config itself is illegal, and the logic in Language.java means the language is legal but not implemented.

xfl03:
> > > You shouldn't delete the handle on "und".
> > 
> > 
> > @iTruth think handle on `und` is unnecessary, so do I.
> > The fallback logic to `en-US` is in `Language.java`.
> 
> und means the locale language config itself is illegal, and the logic in Language.java means the language is legal but not implemented.

There is a fallback logic here: https://github.com/Grasscutters/Grasscutter/blob/stable/src/main/java/emu/grasscutter/utils/Language.java#L52

lhhxxxxx:
> > > > You shouldn't delete the handle on "und".
> > > 
> > > 
> > > @iTruth think handle on `und` is unnecessary, so do I.
> > > The fallback logic to `en-US` is in `Language.java`.
> > 
> > und means the locale language config itself is illegal, and the logic in Language.java means the language is legal but not implemented.
> 
> There is a fallback logic here: https://github.com/Grasscutters/Grasscutter/blob/stable/src/main/java/emu/grasscutter/utils/Language.java#L52

I obviously know there is fallback logic here, but what I mean is that und needs to be handled separately, otherwise the warning on server startup will be "Failed to load language file : und.json", and someone who doesn't understand the source code can't know what wrong there is with the config.

iTruth:
> > You shouldn't delete the handle on "und".
> 
> @iTruth think handle on `und` is unnecessary, so do I. The fallback logic to `en-US` is in `Language.java`.

I'm not saying handle on `und` is unnecessary, contrary, I think it will be helpful for someone who dont't know what is `und` means.  
what I want to say is there is no necessary to fallback to `en-US` in `loadLanguage` and I think handle `und` in Language.java is better

xfl03:
This problem has been fixed.

PR #625, opened 2022-05-07
More reliable stamina calculation

gentlespoon:
by separately handling immediate one-time cost and cost over time.


gentlespoon:
Stamina calculation will now stop on game pause.

ghost (approved):

Issue #183, opened 2022-04-24
Gacha prob and mechanism is not correct


WetABQ:
Gacha prob has prob increasing mechanism and stable mechanism

and UP gacha pool does not contain Amber, Kaeya, and Lisa 

4 star UP also has the pity mechanism.

We could refer to [this passage (Chinese)](https://www.bilibili.com/read/cv10468091)
And I also once wrote a QQ bot Gacha Emulator with an overall customize pipeline system for it, you could refer [this code (write in Kotlin)](https://github.com/StarWishsama/Comet-Bot/tree/dev/src/main/kotlin/io/github/starwishsama/comet/genshin/gacha)

## Pity & UP

![image](https://user-images.githubusercontent.com/25319400/164965634-359888ab-7330-4fc0-9e37-8d37d07d8d5c.png)

## Stable Mechanism

![image](https://user-images.githubusercontent.com/25319400/164965764-18e26eab-40a4-490a-ab2d-659df17223ab.png)

## Overall Pipeline

![image](https://user-images.githubusercontent.com/25319400/164965758-49a19fc5-a5b4-43f5-81b4-6082636c25af.png)

WetABQ:
I will translate it later ;)

lunaticwhat:
since gacha algorithm is still a PoC and hardcoded in some way, we still need to improve it with better math !

Jirehlov:
I'd recommend the same algorithm as well.
For more reference only,
PoC in C++: https://github.com/Jirehlov/libgiwscli
PoC in Python: https://github.com/CMHopeSunshine/Genshin_Gacha_XY

Also some tips,
1. The counters in the "Stable Mechanism" should be cleared once the current banner ends;
2. Mersenne Twister is recommended;
3. Expected probs are,
for weapon banners, 
5-star items overall prob: 1.87793%
4-star items overall prob: 14.8412%
for char banners and Permanent Wish,
5-star items overall prob: 1.60521%
4-star items overall prob: 13.0566%

PR #639, opened 2022-05-07
[Config-breaking] Gacha revamp

Birdulon:
## Key Goals & Concepts
- Use concepts outlined in #183 with default rates for each banner adjusted to better fit observed data from paimon.moe
### No base rates or pity, only lerp arrays
Instead of configuring via base rates, soft/hard pity, define some fixed points to lerp between.
For example, the 3-piece piecewise function for the weapon banner's 5-star probability mechanism shown below can be exactly represented as `"weights5": [[1,70], [62,70], [73,7770], [80,10220]]` in `Banners.json`
image
This is a relatively complex pity mechanism consisting of three different slopes, yet it is trivial to define and edit.
As for why this shouldn't just be hardcoded, statistics show that the 5* rates on all three vanilla banners are considerably different:
image
There's a clear trend on the standard banner - the 5* probability starts at 0.75% and climbs up to 1.5% before the steep climb of soft pity kicks in. The weapon and character banners are relatively flat in this region. The pity slopes on standard and character mostly line up, but weapon does its own thing.
| Banner | Stage 1 | Stage 2 | Stage 3 | Example Lerp |
| --- | --- | --- | --- | --- |
| Standard | Rising | Sharp rise | - | `[[1,75], [73,150], [90,10000]]` |
| Character | Flat | Sharp rise | - | `[[1,80], [73,80], [90,10000]]` |
| Weapon | Flat | Sharp rise | Slow rise | `[[1,70], [62,70], [73,7770], [80,10220]]` |

These would be a pain to hardcode each case for, and we'd be doing no favors to anyone who wants to mod their banner rates. This new format makes it easy for us to specify (current) vanilla behaviour and gives powerful tools to anyone who wants control over their banner rates. An approximation of the old variables in the new format would be `[[1, baseYellowRate], [softPity, baseYellowRate], [hardPity, 10000]]`

### More than just 5-stars
With all that said, as just a mechanism for 5-star rates, this might be an excessive rework. The real issue is that this also digs into 4-star rate behaviour (the weapon banner is different to the other two, peaks on 8 instead of 9), and non-rateUp (termed fallback in this PR) reward pools (avatars and weapons in vanilla, but I generically call them pool1 and pool2 everywhere). There is another rate mechanic similar to pity that is used to ensure that players don't go too long without getting an avatar, or without getting a high-rarity weapon. Some coefficients are proposed in the image below and I've converted a couple of them already.
image

### Better-than-vanilla server configs
I also included a function to remove C6 characters from the reward pools. This currently also preserves guarantees when you "lose a 50:50" by having no rate-up reward available.

### Remaining work
- ~Rigorous testing~
- ~Settling on vanilla default configs that will satisfy #183~
- Deciding on what to do with deprecated config fields. Currently I try to legacy support some of them but it's probably not worth it long-term.
- ~Harden up the edge cases surrounding empty reward pools, especially if removing items from existing pools is a desirable feature.~
- ~More rigorous testing~
- ~Syntax may have gotten a bit ridiculous during some of the refactoring, might need some fixup so that it's actually maintainable~
- ~More rigorous testing after that too~
- Try to stop giving maintainers headaches with ambitious reworks and do something smaller and more manageable next time

### DIFF
--- a/data/Banners.json
+++ b/data/Banners.json
@@ -6,7 +6,9 @@
 		"prefabPath": "GachaShowPanel_A022",
 		"previewPrefabPath": "UI_Tab_GachaShowPanel_A022",
 		"titlePath": "UI_GACHA_SHOW_PANEL_A022_TITLE",
-		"costItem": 224,
+		"costItemId": 224,
+		"costItemAmount": 1,
+		"costItemAmount10": 10,
 		"beginTime": 0,
 		"endTime": 1924992000,
 		"sortId": 1000,
@@ -21,7 +23,7 @@
 		"prefabPath": "GachaShowPanel_A079",
 		"previewPrefabPath": "UI_Tab_GachaShowPanel_A079",
 		"titlePath": "UI_GACHA_SHOW_PANEL_A048_TITLE",
-		"costItem": 223,
+		"costItemId": 223,
 		"beginTime": 0,
 		"endTime": 1924992000,
 		"sortId": 9998,
@@ -37,7 +39,7 @@
 		"prefabPath": "GachaShowPanel_A080",
 		"previewPrefabPath": "UI_Tab_GachaShowPanel_A080",
 		"titlePath": "UI_GACHA_SHOW_PANEL_A021_TITLE",
-		"costItem": 223,
+		"costItemId": 223,
 		"beginTime": 0,
 		"endTime": 1924992000,
 		"sortId": 9997,
@@ -47,6 +49,7 @@
 		"rateUpItems4": [11401, 12402, 13407, 14401, 15401],
 		"rateUpItems5": [11509, 12504],
 		"fallbackItems5Pool1": [],
+		"weights4": [[1,600], [7,600], [8, 6600], [10,12600]],
 		"weights5": [[1,100], [62,100], [73, 7800], [80,10000]]
 	}
 ]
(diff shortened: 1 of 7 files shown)


Birdulon:
Now depends on #824

Birdulon:
Rebased to development.

mcxiaoke:
anyone to merge this?

ghost (approved):

battleTrainer:
那么该如何奖池编辑器呢

ELiuven:
Sorry, may I ask which destination folder should I put the .json file generated by the editor?

Birdulon:
> Sorry, may I ask which destination folder should I put the .json file generated by the editor?

Under `data`
image

Audience03:
生成卡池的json后，应该怎么解析使卡池显示且可使用？

curryyinxh:
如何使用奖池编辑器？

curryyinxh:
> 生成卡池的json后，应该怎么解析使卡池显示并且可以使用？

编辑banners.json文件

987694341:
生成的json编辑banners.json怎么去编辑放置生成的json

CrayyanG:
我想实现每抽都是五星，应该如何修改呢

Leego000:
游戏里不显示奖池 商店该怎么弄呢

K-Summer:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

PR #643, opened 2022-05-07
use default language file

zhaodice:
if language file doesn't exist, try default file


KingRainbow44:
the process of fallback languages has been simplified in `330427f5a56e01ecc494db4e042edad17f1eec40`

PR #645, opened 2022-05-07
Add language support more gracefully

cfuncode:
Snipaste_2022-05-08_01-18-14

Snipaste_2022-05-08_01-19-07

I can't fix some garbled characters. It seems to be logback problem.


cfuncode:
好像已经不需要了。。

PR #659, opened 2022-05-08
Add command description multilingual

577fkj:
## Description

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/Command.java
+++ b/src/main/java/emu/grasscutter/command/Command.java
@@ -9,7 +9,7 @@ public @interface Command {
 
     String usage() default "No usage specified";
 
-    String description() default "No description specified";
+    String description() default "commands.generic.no_description_specified";
 
     String[] aliases() default {};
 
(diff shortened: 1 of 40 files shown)


src/main/java/emu/grasscutter/command/Command.java
```diff
@@ -9,8 +9,6 @@
 
     String usage() default "No usage specified";
 
-    String description() default "No description specified";
```
4Benj, line 12: Instead of removing this you can make the description take the translation key instead

577fkj: OK, I'll change it back

577fkj: I added it back

ghost (changes requested):
The description annotation should be setting the translation key, which you then retrieve with a description method to translate, instead of creating a description override for every command.

577fkj:
> The description annotation should be setting the translation key, which you then retrieve with a description method to translate, instead of creating a description override for every command.

Changed as required

ghost (approved):

PR #688, opened 2022-05-08
Improve logging pattern by including caller class name

ghost:
## Description

Change logging pattern to include caller class name. This change means that contributors no longer need to remember to manually add prefix for messages they log.

Console pattern is now `13:50:09 <INFO:GameServer> Game Server started on port 22102`, while logfile pattern is `2022-05-08T13:50:09Z <INFO:emu.grasscutter.server.game.GameServer> Game Server started on port 22102`.

- Caller class name lookup incurs additional overhead, however since we are not planning on logging tens of thousands messages per second I do not think there is any significance of that.
- Since pattern now includes shorthand class name, I've tried to adjust it to minimize redundant characters that would artifically make log messages longer. As such, `[ ]` wrapping around time (date) has been removed, while level has been merged together with class name and wrapping changed to `< >`.
- While console uses short-hand version of class name, logfile contains full one.
- Logfile pattern has been adjusted to mimic that of console, specifically adding verbosity level which was previously missing.

## Issues fixed by this PR

N/A

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ dependencies {
 
     implementation group: 'io.netty', name: 'netty-all', version: '4.1.71.Final'
 
-    implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.8'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.9.0'
     implementation group: 'com.google.protobuf', name: 'protobuf-java', version: '3.18.2'
 
     implementation group: 'org.reflections', name: 'reflections', version: '0.10.2'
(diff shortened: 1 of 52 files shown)


ghost:
Not a fan of the long log format to be honest, but lets see what the others think about it.

ghost:
By long format you mean one in logfiles? It can be shortened but I thought that it won't make much difference for logfiles and it can help in some rare cases when shortened class name is ambigious.

gentlespoon:
Is it going to print the FQCN or just the CN? Personally I like cn, but fq is probably too much. With cn it is usually good enough to locate the source of the log.

ghost:
At the moment it shows CN (ie. `GameServer`) in console and FQCN in logfiles (ie. `emu.grasscutter.server.game.GameServer`).

KingRainbow44 (approved):

PR #692, opened 2022-05-08
Relics can be added using the give all command

cfuncode:
When using the give all command, added the maximum level of relic for the corresponding star.
However, there are many values after the secondary attribute upgrade that do not belong to the normal range.
And don't know how to fix it.

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/ClearCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/ClearCommand.java
@@ -42,7 +42,8 @@ public final class ClearCommand implements CommandHandler {
             case "art" -> {
             	toDelete = playerInventory.getItems().values().stream()
                         .filter(item -> item.getItemType() == ItemType.ITEM_RELIQUARY)
-                        .filter(item -> item.getLevel() == 1 && item.getExp() == 0)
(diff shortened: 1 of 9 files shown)


ghost:
Related, feature previously removed by https://github.com/Grasscutters/Grasscutter/pull/365

src/main/java/emu/grasscutter/command/commands/ClearCommand.java
```diff
@@ -42,7 +42,8 @@ public void execute(Player sender, Player targetPlayer, List<String> args) {
             case "art" -> {
             	toDelete = playerInventory.getItems().values().stream()
                         .filter(item -> item.getItemType() == ItemType.ITEM_RELIQUARY)
-                        .filter(item -> item.getLevel() == 1 && item.getExp() == 0)
```
ghost, line 45: Note this changes logic of `clear`

cfuncode: Isn't this the right change? Only sacred relics worn and locked will not be deleted. However, the Readme. md content in English has not been corrected.

cfuncode:
> Related, feature previously removed by #365

It wasn't removed, was it? Isn't it still working?

PR #704, opened 2022-05-08
Added to ability to specify main and substats for /giveart via names instead of IDs

GanyusLeftHorn:
Allow the use of nice names for specifying main and substats when using `/giveart`, because I don't want to look up IDs all the time. For example, the following commands are now valid:

```
/giveart 77524 atk cr_4,3 cdmg_4,3 atk_4 21
/giveart 59530 cr cdmg_4,3 cdmg_3 er_2,2 21
/giveart 63110 phys% 1
/giveart 63410 cryo% cr_3,10 atk%_4 cdmg_4 er_4 21
/giveart 73550 er em_4,3 cr_3,2 cdmg_3 cdmg_4 atk_4,2 21
```

Main stats are specified as one of `hp`, `hp%`, `atk`, `atk%`, `def%`, `er`, `em`, `hb`, `cdmg`, `cr`, `phys%`, `dendro%`, `geo%`, `anemo%`, `hydro%`, `cryo%`, `electro%` and `pyro%`. The implementation is aware that mainstat IDs change depending on artifact piece, so for example `atk%` is mapped to `10990` for sands, `50990` for goblets and `30990` for circlets. The mapping only contains legal main stats for the given piece, so for example `/giveart 73550 cr em_4,3 cr_3,2 cdmg_3 cdmg_4 atk_4,2 21` gives an error (sands can't have a crit rate main stat).

Substats are specified as one of `hp`, `hp%`, `atk`, `atk%`, `def`, `def%`, `er`, `em`, `cr` and `cdmg`, and are followed by a number indicating the roll tier. So, for example, a max roll crit rate substat is `cr_4` (tier 4 roll). The implementation selects the right substat ID based on the specified stat, tier, as well as the rarity of the artifact (the commonly used substat IDs follow the pattern `RSSSST`, where `R` = rarity, `SSSS` = 4 digits specifying the stat, and `T` = tier). The implementation checks whether the specified tier is valid for the given piece. For example, `/giveart 51210 hp% atk_4 21` gives an error because a 2-star piece cannot have a tier 4 substat.

Of course, specifying numeric IDs instead of names is still possible as before.

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/GiveArtifactCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/GiveArtifactCommand.java
@@ -70,14 +70,26 @@ public final class GiveArtifactCommand implements CommandHandler {
 
 		// If the argument was not an integer, we try to determine
 		// the append prop ID from the given text + artifact information.
-		// A substat string has the format `substat_tier`.
+		// A substat string has the format `substat_tier`, with the
+		// `_tier` part being optional.
 		String[] substatArgs = substatText.split("_");
-		if (substatArgs.length != 2) {
+		String substatType;
+		int substatTier;
+
+		if (substatArgs.length == 1) {
+			substatType = substatArgs[0];
+			substatTier = 
+				itemData.getRankLevel() == 1 ? 2 
+				: itemData.getRankLevel() == 2 ? 3 
+				: 4;
+		}
+		else if (substatArgs.length == 2) {
+			substatType = substatArgs[0];
+			substatTier = Integer.parseInt(substatArgs[1]);
+		}
+		else {
 			throw new IllegalArgumentException();
 		}
-
-		String substatType = substatArgs[0];
-		int substatTier = Integer.parseInt(substatArgs[1]);
 
 		// Check if the specified tier is legal for the artifact rarity.
 		if (substatTier < 1 || substatTier > 4) {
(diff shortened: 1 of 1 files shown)


ghost:
We do have quite a lot of community tools for generating artifact commands, why not use these?

>gives an error (sands can't have a crit rate main stat)

They don't exist on vanilla but they certainly can exist in-game, in a same way as plume can have HP instead of ATK. Is there any reason for artificially limiting it for GM commands?

Where such mapping (for legal stats/substat) would be more useful is in drop logic however, where stats are generated.

ghost (approved):

GanyusLeftHorn:
Small update: I made the `_tier` suffix optional. The implementation will default to the highest possible tier given the piece's rarity. So, for example, `/giveart 77524 atk cr` ir euqivalent to `/giveart 77524 atk cr_4`.

> They don't exist on vanilla but they certainly can exist in-game, in a same way as plume can have HP instead of ATK. Is there any reason for artificially limiting it for GM commands?

Yep, they certainly exist, and you can still add them by simply using the ID directly. But for a name->ID mapping, I'm honestly not sure how to handle them completely. For example, take ATK%, for which we have three different IDs (`10990` for sands, `50990` for goblets and `30990` for circlets). When someone now specifies `atk%` for a flower or a feather, which one of these IDs do we use?

ghost:
>When someone now specifies atk% for a flower or a feather, which one of these IDs do we use?

Indeed, that shows the issue.

I still don't think this will be particularly useful given it still needs initial artifact ID to be specified, people will just use command generators for that, be it with this PR or without, but I have nothing against it either and will most likely use it myself to quickly give artifacts.

PR #751, opened 2022-05-09
refactor command map

Qiuk17:
## Description

**Please let me know if this can be helpful.** @KingRainbow44 
And forgive me for the awful commit logs.

### Current shortages

- A bunch of repeated codes to deal with optional arguments and invalid user input.
- Command parsing and what the handlers meant to do are tightly coupled.

### Advantages after being refactored

- Multiple line input and prompt indicator.
- Elegent and simple `Commands`.
- Plugins can reuse the handlers instead of reimplementing one. (maybe this can help with `gcgm-plugin` by @4Benj )
- The extendable design makes it easier to implement other command clients such as an embedded web shell (based on `xterm.js` and websocket) in any web page in the game.

#### Before
```java
// AccountCommand.java
@Command(label = "account", usage = "account <create|delete> <username> [uid]", description = "commands.account.description")
public final class AccountCommand implements CommandHandler {

    @Override
        if (sender != null) {
            CommandHandler.sendMessage(sender, translate("commands.generic.console_execute_error"));
            return;
        }

        if (args.size() < 2) {
            CommandHandler.sendMessage(null, translate("commands.account.command_usage"));
            return;
        }

        String action = args.get(0);
        String username = args.get(1);

        switch (action) {
            default:
                CommandHandler.sendMessage(null, translate("commands.account.command_usage"));
                return;
            case "create":
                int uid = 0;
                if (args.size() > 2) {
                    try {
                        uid = Integer.parseInt(args.get(2));
                    } catch (NumberFormatException ignored) {
                        CommandHandler.sendMessage(null, translate("commands.account.invalid"));
                        return;
                    }
                }

                emu.grasscutter.game.Account account = DatabaseHelper.createAccountWithId(username, uid);
                if (account == null) {
                    CommandHandler.sendMessage(null, translate("commands.account.exists"));
                    return;
                } else {
                    account.addPermission("*");
                    account.save(); // Save account to database.

                    CommandHandler.sendMessage(null, translate("commands.account.create", Integer.toString(account.getPlayerUid())));
                }
                return;
            case "delete":
                if (DatabaseHelper.deleteAccount(username)) {
                    CommandHandler.sendMessage(null, translate("commands.account.delete"));
                } else {
                    CommandHandler.sendMessage(null, translate("commands.account.no_account"));
                }
        }
    }
}
```

#### After
```java
// AccountCommand.java
@Command(literal = "account")
public class AccountCommand {
    @SubCommandHandler("create")
    @AllowedOrigin(Origin.SERVER)
    @Description("commands.account.description.create")
    public void createAccount(BaseCommandSource source, String username, @OptionalArgument Integer uid) {
        CommandParser.dispatch(
                source,
                AccountHandlerCollection.ACCOUNT_CREATE,
                HandlerContext.builder()
                        .content(AccountHandlerCollection.Fields.ACCOUNT_USERNAME, username)
                        .content(AccountHandlerCollection.Fields.ACCOUNT_UID, uid)
        );
    }

    @SubCommandHandler("delete")
    @AllowedOrigin(Origin.SERVER)
    @Description("commands.account.description.delete")
    public void deleteAccount(BaseCommandSource source, String username) {
        CommandParser.dispatch(
                source,
                AccountHandlerCollection.ACCOUNT_DELETE,
                HandlerContext.builder()
                        .content(AccountHandlerCollection.Fields.ACCOUNT_USERNAME, username)
        );
    }
}
```
```java
// AccountHandlerCollection.java
@HandlerCollection
public final class AccountHandlerCollection {
    @Handler(ACCOUNT_CREATE)
    @SneakyThrows
    public void createAccount(HandlerContext context) {
        String username = context.getRequired(Fields.ACCOUNT_USERNAME, String.class);
        int uid = context.getOptional(Fields.ACCOUNT_UID, 0);
        Account account = DatabaseHelper.createAccountWithId(username, uid);

        if (account == null) {
            context.notify(translate("commands.account.exists"));
            return;
        }
        account.addPermission("*");
        account.save();
        context.notify(translate("commands.account.create", Integer.toString(account.getPlayerUid())));
    }

    @Handler(ACCOUNT_DELETE)
    @SneakyThrows
    public void deleteAccount(HandlerContext context) {
        String username = context.getRequired(Fields.ACCOUNT_USERNAME, String.class);
        if (DatabaseHelper.deleteAccount(username)) {
            context.notify(translate("commands.account.delete"));
        } else {
            context.notify(translate("commands.account.no_account"));
        }
    }

    public static final String ACCOUNT_CREATE = "account.create";
    public static final String ACCOUNT_DELETE = "account.delete";

    public static final class Fields {
        public static final String ACCOUNT_USERNAME = "account.username";
        public static final String ACCOUNT_UID = "account.uid";
    }
}
```

### Demo for multiple line input

mail
mail

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


KingRainbow44:
AMAZING. Is there a way you can add a method for taking in a locale and returning the translated string instead of having to call `Language#translate`? (and keep the method for raw strings ofc)

src/main/java/emu/grasscutter/command/source/impl/ClientChatCommandSource.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public HandlerContext buildContext(HandlerContext.HandlerContextBuilder contextBuilder) {
+        HandlerContext context = contextBuilder
+                .errorConsumer(e -> error(e.toString()))
+                .resultConsumer(r -> info(r != null ? "Handler result: %s.".formatted(r.toString()) : "Handler completed."))
+                .messageConsumer(m -> info(m.toString()))
+                .build();
+        if (context.getContent().get(ContextFields.TARGET_UID) == null) {
+            Integer targetUid = (Integer) get(PERSISTED_TARGET_KEY);
+            if (targetUid != null) {
```
Birdulon: Does this default to the in-game sender elsewhere?

Qiuk17: In both `ChatManager#sendPrivateMessage` and `ChatManager#sendPrivateMessage`

Qiuk17:
> AMAZING. Is there a way you can add a method for taking in a locale and returning the translated string instead of having to call `Language#translate`? (and keep the method for raw strings ofc)

I will try.

src/main/java/emu/grasscutter/command/source/impl/ClientChatCommandSource.java
Birdulon: Currently target priority works like this:

1. @UID as an argument inside command
2. Recipient of message in in-game chat
3. Persistent target set by `/target`
4. Sender

> In both `ChatManager#sendPrivateMessage` and `ChatManager#sendPrivateMessage`

That just calls `ClientChatCommandSource.getCommandSourceByPlayer(player).runCommand(message.substring(1));`, how does it handle 2. or 4.?

src/main/java/emu/grasscutter/command/handler/HandlerContext.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    /**
+     * <p>Pass the result so that others can display.</p>
+     * <b>Notice: </b><p>Code under this will NOT be executed.</p>
+     * @param result the result of this handler
+     */
+    public void returnWith(Object result) throws HandlerSuccess {
+        throw new HandlerSuccess(result);
```
jie65535: I don't understand the purpose here, why throw `'Exception'` of `HandlerSuccess`?

src/main/java/emu/grasscutter/command/source/impl/ClientChatCommandSource.java
Qiuk17: '4.' can be handled by overriding `ClientChatSource#buildContext`.
'2.' can be handled by override `runCommand`. add a parameter `recipient`, save it to `persistedContext`, then call `super.runCommand`

Qiuk17: "@uid" as an argument cannot be parsed now. you call tell the program how to deal with it in `ParseUtil.PermittedClasses`

src/main/java/emu/grasscutter/command/handler/HandlerContext.java
Qiuk17: `HandlerSuccess` is not an exception. It just interrupts the code and gets caught in `DispatchListener#dispatch`. So that the consumer can be acknowledged that the handler is DONE and you can dispose resources in all callbacks

jie65535: I noticed that your example usage is called after handler is DONE, what's the difference between letting the method return directly?

src/main/java/emu/grasscutter/command/source/impl/ClientChatCommandSource.java
Birdulon: > '4.' can be handled by overriding `ClientChatSource#buildContext`. '2.' can be handled by override `runCommand`. add a parameter `recipient`, save it to `persistedContext`, then call `super.runCommand`

> @uid as an argument cannot be parsed now. you call tell the program how to deal with it in `ParseUtil.PermittedClasses`

Ok, can you add those to your PR?

src/main/java/emu/grasscutter/command/handler/HandlerContext.java
Qiuk17: In my opinion, handlers can return something like "online players". `context#notify` just tells the consumers about the progresses etc. It is ok to just access to `HandlerContext#resultConsumer` an then return. But I don't recommand to do so because I want it to seem like "returning something" in a void method.

src/main/java/emu/grasscutter/command/source/impl/ClientChatCommandSource.java
Qiuk17: of course yes. and maybe we can ask for a new branch so we can work together.

src/main/java/emu/grasscutter/command/handler/HandlerContext.java
jie65535: > The signature of %s must be `void (HandlerContext)`.

Maybe make the signature have a return value?
 `void (HandlerContext)` ->  `object (HandlerContext)`

Or put the returned result into the context.

Because throwing HandleSuccess feels weird...

Qiuk17:
@KingRainbow44 
Is it ok for us to have another branch so we can keep developing this without borthering development builds?

src/main/java/emu/grasscutter/command/handler/HandlerContext.java
Qiuk17: Oh yes. In my previous code the handlers are managed by eventbus so they are forced to be void. Now it can just return.
silly me.

src/main/java/emu/grasscutter/command/source/impl/ClientChatCommandSource.java
Qiuk17: > 1. @uid as an argument inside command

Done. and custom argument types are now supported.

lhhxxxxx:
Is this still in working?
Some part of the feature I'd like to introduce may require the use of some methods in CommandMap.

Qiuk17:
> Is this still in working? Some part of the feature I'd like to introduce may require the use of some methods in CommandMap.

yes. but i have to prepare for my finals.

lhhxxxxx:
> > Is this still in working? Some part of the feature I'd like to introduce may require the use of some methods in CommandMap.
> 
> yes. but i have to prepare for my finals.

Then I'll wait.

PR #753, opened 2022-05-10
Add one key graduate (/graduate)

nigpo:
## Description
Add one key graduate for import artifact into bag.
Fix NPE.
Support Android Studio for compile and run.
Support Termux (JDK11) for compile and run.
Fix source code could not be compiled into Termux, now this project can run over an Android phone via Termux.

## Issues fixed by this PR
- [ x ] Bug fix
- [ x ] New feature 
- [ x ] Enhancement
- [ ] Documentation

## Checklist:

- [ x ] My code follows the style guidelines of this project
- [ x ] My pull request is unique and no other pull requests have been opened for these changes
- [ x ] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [ x ] I am responsible for any copyright issues with my code if it occurs in the future.


src/main/java/emu/grasscutter/scripts/serializer/LuaSerializer.java
```diff
@@ -76,10 +76,7 @@ public <T> T serialize(Class<T> type, LuaTable table) {
 			for (LuaValue k : keys) {
 				try {
 					Field field = object.getClass().getDeclaredField(k.checkjstring());
-					if (field == null) {
```
ghost, line 79: Please explain this change

src/main/java/emu/grasscutter/data/def/ShopGoodsData.java
```diff
@@ -36,12 +36,13 @@ public void onLoad() {
        if (this.RefreshType == null)
            this.RefreshTypeEnum = ShopInfo.ShopRefreshType.NONE;
        else {
-           this.RefreshTypeEnum = switch (this.RefreshType) {
-               case "SHOP_REFRESH_DAILY" -> ShopInfo.ShopRefreshType.SHOP_REFRESH_DAILY;
-               case "SHOP_REFRESH_WEEKLY" -> ShopInfo.ShopRefreshType.SHOP_REFRESH_WEEKLY;
-               case "SHOP_REFRESH_MONTHLY" -> ShopInfo.ShopRefreshType.SHOP_REFRESH_MONTHLY;
-               default -> ShopInfo.ShopRefreshType.NONE;
-           };
+//           this.RefreshTypeEnum = null;
```
ghost, line 39: Unnecessary comment?

build.gradle
```diff
@@ -8,6 +8,13 @@
 
 buildscript {
     repositories {
+        maven{ url 'https://maven.aliyun.com/nexus/content/repositories/central/'}
```
ghost, line 11: Do we really need all of these?

src/main/java/emu/grasscutter/command/commands/GraduateCommand.java
```diff
@@ -0,0 +1,160 @@
...
+            {
+                toWho = sender;
+            }
+            else
+            {
+                CommandHandler.sendMessage(sender, Language.translate("commands.execution.need_target"));
+                return;
+            }
+            String[][] graduateData = new String[][]{
+// 神里绫人毕业圣遗物
```
ghost: Please use English in comments if possible

src/main/java/emu/grasscutter/command/commands/GiveAllCommand.java
```diff
@@ -22,7 +22,8 @@ public final class GiveAllCommand implements CommandHandler {
     public void execute(Player sender, Player targetPlayer, List<String> args) {
         if (targetPlayer == null) {
             CommandHandler.sendMessage(sender, translate("commands.execution.need_target"));
-            return;
+//            return;
+            targetPlayer = sender;
```
ghost: This is incorrect behavior as per new targeting system. You are also being inconsistent with both displaying error message AND continuing to execute.

ghost (changes requested):

src/main/java/emu/grasscutter/game/shop/ShopManager.java
```diff
@@ -19,6 +19,8 @@
 public class ShopManager {
 	private final GameServer server;
 
+	public static Boolean DEBUG_ULTIMATED_SHOP = Boolean.FALSE;
```
ghost, line 22: What is the purpose of this? You can always just give yourself enough resources to buy items.

src/main/java/emu/grasscutter/game/sword/Sword.java
```diff
@@ -0,0 +1,5 @@
+package emu.grasscutter.game.sword;
+
+public class Sword {
+    //TODO: Add sword to world.
```
ghost, line 4: No idea what `Sword` and `SwordManager` here refer to but this is just placeholder boilerplate and definitely doesn't belong in this PR.

src/main/java/emu/grasscutter/server/packet/send/PacketSceneTeamUpdateNotify.java
```diff
@@ -24,14 +25,19 @@ public PacketSceneTeamUpdateNotify(Player player) {
...
-						.setWeaponGuid(entityAvatar.getAvatar().getWeapon().getGuid())
+//						.setWeaponGuid(entityAvatar.getAvatar().getWeapon().getGuid())
 						.setWeaponEntityId(entityAvatar.getWeaponEntityId())
 						.setIsPlayerCurAvatar(p.getTeamManager().getCurrentAvatarEntity() == entityAvatar)
 						.setIsOnScene(p.getTeamManager().getCurrentAvatarEntity() == entityAvatar)
 						.setAvatarAbilityInfo(AbilitySyncStateInfo.newBuilder())
 						.setWeaponAbilityInfo(AbilitySyncStateInfo.newBuilder())
 						.setAbilityControlBlock(entityAvatar.getAbilityControlBlock());
-				
+				GameItem weapon = entityAvatar.getAvatar().getWeapon();
```
ghost, line 35: Please explain this change too. Under what circumstances does this "crash" occur and under what circumstances would avatar weapon be `null`?

build.gradle
```diff
@@ -46,14 +46,14 @@ plugins {
 compileJava.options.encoding = "UTF-8"
 compileTestJava.options.encoding = "UTF-8"
 
-sourceCompatibility = JavaVersion.VERSION_17
-targetCompatibility = JavaVersion.VERSION_17
+sourceCompatibility = JavaVersion.VERSION_11
```
ghost, line 49: Not sure we would like to get back to Java 11. At best, I'd recommend 16.

src/main/java/emu/grasscutter/scripts/serializer/LuaSerializer.java
nigpo: Field should always non-null.

build.gradle
ghost: Also going back to Java 11 isn't strictly necessary for this to work on Termux or be compiled on Android, you can just install `openjdk-17`.

build.gradle
nigpo: You can import dependency quickly after import all of those.

src/main/java/emu/grasscutter/command/commands/GiveAllCommand.java
nigpo: Give all command will detect if sender is valid, so don't mind if sender is valid.

build.gradle
nigpo: Android Studio not support Java 16, Termux too.

ghost: I tend to disagree.

Termux has official package for OpenJDK 17 - https://github.com/termux/termux-packages/tree/master/packages/openjdk-17

Also this server is not meant to be built by Android Studio.

nigpo: ARM64 platform have no way for compile Java 17 code, and no way for build this project if set Java version to 17.
Android Studio does not support Java 16 currently, you can ask Google if Android 14 support Java17 or not.

src/main/java/emu/grasscutter/command/commands/GiveAllCommand.java
ghost: Please see https://github.com/Grasscutters/Grasscutter/pull/499. You are breaking behavior of the command system by this change and not even doing it properly, by both continuing execution AND displaying (now improper) error message.

build.gradle
nigpo: This is why AIDE or AndroidIDE only support Java11 and not support kotlin or Java17.

src/main/java/emu/grasscutter/command/commands/GiveAllCommand.java
ghost: In addition, this should NOT be necessary - `CommandMap` should handle translation of current player to `targetPlayer` if no other target was provided, without separate commands having to deal with it.

build.gradle
ghost: I understand that these are local CDNs for faster download, however we do not need all of these. If you really want a proper CDN then you should simply use `https://maven.aliyun.com/repository/gradle-plugin`, as that is CDN for the one we currently use (`https://plugins.gradle.org/m2/`).

Also Google ones contain just Android SDK stuff, we do not need those.

build.gradle
ghost: OpenJDK 17 certainly does work on ARM64. Not sure we want to artificially block this project from benefiting from years of Java development simply so someone can build it on Android.

src/main/java/emu/grasscutter/command/commands/GiveAllCommand.java
nigpo: This changes revert after deal with a bug of NPE, and it will be nothing happened after both of sender is null or empty.

build.gradle
ghost: Here's direct link for OpenJDK 17 for Termux on aarch64 (arm64): https://packages.termux.org/apt/termux-main/pool/main/o/openjdk-17/openjdk-17_17.0-21_aarch64.deb

Since that is a full JDK, I see no reason why you could not run server that way. As far as I see, the only thing you're losing by not supporting 11 would be inability to bundle this server in Android app as all-in-one APK file, which we are not interested in supporting.

src/main/java/emu/grasscutter/command/commands/GraduateCommand.java
nigpo: Translation was updated

build.gradle
nigpo: https://androidide.com/
There is no way for compile code on Android device before downgrade to Java11, after downgrade, those code can compile on Android device or Android Studio. After compile those code, you can launch it or upload it to server without recompile.

nigpo: If you wanna enjoy newest technology or features of language, I think that you should transfer this project into kotlin even php8, I am delight to see this server can run over php server which can reduce cost if you public it to internet.

Birdulon:
Did you run a bunch of unrelated files into a linter that swaps arrow cases seemingly at random?

EDIT: Oh I see, it's to remove all Java 12+ features.

mcxiaoke:
this function should be supplied as external tool, eg a shell script or just copy paste command list txt

ghost:
Thank you for making a pull request. However, I'm not sure what some of these changes are supposed to be about, specifically the Sword class you added. Also, we've just migrated over to java 17 and will probably not be downgrading to 11 without a good reason1.

Scirese:
In fact now we can compile and run the server under a chrooted Linux env on a rooted phone,and by this way we can bacically got the same exprenience of deploying it on a linux computer. If your phone is not rooted,like you said,Termux could handle all the thing.And just like HotaruYS said, Termux has a full JDK17 package.Its not hard to make grasscutter runs in termux,The biggest problem is to make mongodb runs properly under termux(lets assume that u wants to run all of the things in a android phone).And really,I dont think we need to compile it in android studio or ide,so I think we dont have to downgrade to java11.

mingjun97:
I guess your command can be implemented as a grasscutter plugin.

PR #797, opened 2022-05-11
Introduce `-version` option to display version

mingjun97:
## Description

* Generate a file that holds git hash and version at build time.
* Add option that can help print out the version info. This could benifit manage the versioning stuffs in `Issues`.

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,7 +122,7 @@ jar {
     from('src/main/java') {
         include '*.xml'
     }
-    new File(projectDir, "src/generated/main/java/emu/grasscutter/BuildConfig.java").text = """
+    new File(projectDir, "src/main/java/emu/grasscutter/BuildConfig.java").text = """
     package emu.grasscutter;
     public class BuildConfig {
             public static final String VERSION = \"${version}\";
--- a/build.gradle
+++ b/build.gradle
@@ -45,13 +45,6 @@ targetCompatibility = JavaVersion.VERSION_17
 group = 'xyz.grasscutters'
 version = '1.1.1-dev'
 
-def gitCommitHash = {
-    try {
-        return 'git rev-parse --verify --short HEAD'.execute().text.trim()
-    } catch (e) {
-        return "GIT_NOT_FOUND"
-    }
-}
 
 sourceCompatibility = 17
 targetCompatibility = 17
@@ -122,13 +115,6 @@ jar {
     from('src/main/java') {
         include '*.xml'
     }
-    new File(projectDir, "src/main/java/emu/grasscutter/BuildConfig.java").text = """
-    package emu.grasscutter;
-    public class BuildConfig {
-            public static final String VERSION = \"${version}\";
-            public static final String GIT_HASH = \"${gitCommitHash()}\";
-    }
-    """
 
     destinationDir = file(".")
 }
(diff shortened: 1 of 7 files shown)


mingjun97:
Well.. Test locally and it works fine. But github action reject the changes. I'll play with it to figure out a solution.

mcxiaoke:
also add this to server start log message better , and append to the output jar name

src/main/java/emu/grasscutter/Grasscutter.java
```diff
@@ -79,6 +80,9 @@ public static void main(String[] args) throws Exception {
 				case "-gachamap" -> {
 					Tools.createGachaMapping(Grasscutter.getConfig().DATA_FOLDER + "/gacha_mappings.js"); exitEarly = true;
 				}
+				case "-version" -> {
```
4Benj, line 83: Why do we need a version argument?

src/main/resources/languages/en-US.json
```diff
@@ -45,7 +45,8 @@
       "run_mode_error": "Invalid server run mode: %s.",
       "run_mode_help": "Server run mode must be 'HYBRID', 'DISPATCH_ONLY', or 'GAME_ONLY'. Unable to start Grasscutter...",
       "create_resources": "Creating resources folder...",
-      "resources_error": "Place a copy of 'BinOutput' and 'ExcelBinOutput' in the resources folder."
+      "resources_error": "Place a copy of 'BinOutput' and 'ExcelBinOutput' in the resources folder.",
+      "version": "Grasscutter version: %s, Git Hash: %s"
```
4Benj: Recommended layout 1.1.0-1234a88
Don’t seperate the version and hash

build.gradle
```diff
@@ -113,6 +122,13 @@ jar {
     from('src/main/java') {
         include '*.xml'
     }
+    new File(projectDir, "src/main/java/emu/grasscutter/BuildConfig.java").text = """
```
4Benj: Maybe something like [this](https://stackoverflow.com/a/33026919) will be better?

4Benj (changes requested):

src/main/java/emu/grasscutter/Grasscutter.java
mingjun97: Magix mentioned we probably want to ask user put the version inside the issue. A version argument should be relatively easy for users to retrieve the version information.

4Benj: I feel like the version should be printed/logged on start up instead. If it already is, then I think this is unneeded redundancy.

build.gradle
mingjun97: I'll try to implement it in this way.

src/main/java/emu/grasscutter/Grasscutter.java
mingjun97: I also implemented printed the version out at this commit. [5e74897](https://github.com/Grasscutters/Grasscutter/pull/797/commits/5e74897b994435b20e788b46507eaacaa2917831)

I can remove the `-version` argument if you like. I still think it's useful in some cases though.

build.gradle
mingjun97: Fixed in the [5aefad6](https://github.com/Grasscutters/Grasscutter/pull/797/commits/5aefad61eacfad17c46a4f139b5d5c8f56c0aff7).

4Benj (approved):

ghost (approved):

PR #821, opened 2022-05-11
Fix Expedition

XTL676:
## Description
*  make expeditionCountLimit change by player's adventure level
*  add avatar's expedition effect(such as -25% expedition time)
## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/expedition/ExpeditionInfo.java
+++ b/src/main/java/emu/grasscutter/game/expedition/ExpeditionInfo.java
@@ -37,8 +37,17 @@ public class ExpeditionInfo {
         this.startTime = startTime;
     }
 
+    public float getShortenRatio() {
+        return shortenRatio;
+    }
+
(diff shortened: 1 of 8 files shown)


ShigemoriHakura (commented):
In PacketAvatarExpeditionCallBackRsp.java I think there's NO need to check if getShortenRatio == 1
In PacketAvatarExpeditionAllDataRsp.java As the limit is increased by level, I think you can just simplify the determine function
In HandlerAvatarExpeditionStartReq.java and HandlerAvatarExpeditionGetRewardReq.java , I think the ratio and avaterId should be saved in json file instead of hard-coded.

ghost (changes requested):
Talent ids should not be hardcoded at all. In addition, you added expedition limits by player level, but you did not update expedition limits when the player levels up.

Issue #664, opened 2022-05-08
Possible to emulate request to version.json + possible steps for full offline support?


shogunator:
Related: #326 

I'm trying to get Grasscutter + game client to work completely offline.

So far the steps I've taken are:

- Run a local DNS server (https://maradns.samiam.org/)
- Configure the ipv4 names similar to proxy.py (https://github.com/Grasscutters/Grasscutter/blob/a185cd167ea38a61fcc0aac1d286377cb89e335c/proxy.py)
- Add a virtual network adapter in Windows, set the DNS server to 127.0.0.1 so MaraDNS is able to resolve the requests

At this point the client opens and does not give me a 4201 error anymore even without an internet connection, however it is now stuck at "Checking for updates" then times out.

Checking the log files in AppData/LocalLow for the game show a request to:

https://webstatic-sea.hoyoverse.com/admin/mi18n/plat_oversea/m2020030410/m2020030410-version.json

Response is:

`{"version":65}`

The error in the log file is just `error=Generic/unknown HTTP error`.

---

Checking if Grasscutter has that route defined in a browser + DevTools, it simply gives the 404 not found page instead of the JSON.

---

Edit: it seems that the server has a version hardcoded in the code. Is it possible to make this configurable?

Relevant line: https://github.com/Grasscutters/Grasscutter/commit/582d91227cf14d729954e8c78a919e265913cdef

I added a route for it on my local repository & build, and got past the initial error 4201 (unable to connect)

---

Edit 2: there's also a HTTP (not HTTPS) request when the game tries to download an update.

```
{"remoteName": "blocks/00/22551915.blk", "md5": "0cf3d6b599d443c7cfa4a3e6189a7757", "fileSize": 3184}
{"remoteName": "blocks/00/25060239.blk", "md5": "f3cf18d697e3380b2f833ccf2c7d4194", "fileSize": 3328}
{"remoteName": "blocks/00/29342328.blk", "md5": "bc652abe9338f1b94f112ad5f8a2e16d", "fileSize": 11245}
{"remoteName": "blocks/00/32070509.blk", "md5": "4bf2a81d7565301269d1c90e14393045", "fileSize": 698}
{"remoteName": "blocks/00/33067900.blk", "md5": "0b4d781c1633537d2fa4b4f0f567d2ca", "fileSize": 454}
```

http://autopatchhk.yuanshen.com/client_design_data/2.6_live/output_6731353_adfeb8be71/client_silence/General/AssetBundles/data_versions

[...]

Issue #326, opened 2022-04-28
Full offline support maybe 


Pull256:
When I disconnected the internet, the game will display an error 4201, which means "No network connection".
I think the proxy setting does not cut down all connections with mihoyo, may be the game itself need internet connection to mihoyo for update?
If possible, it should be possible to remove the network detection in some way to prevent the game from updating or chanting Mihayo to find out what we are doing.

Thanks for all the developers of this software.

memetrollsXD:
The server should already support having no internet, are you sure you redirected all traffic?

Pull256:
I just make a small test with fidder to find out why it happened, here are my result.
The fidder script I use have 24 rules downloaded from github.
When I start the game and enter the login screen, the game send many request to _log-upload-os.mihoyo.com:443_. But when the internet is disconnected, the DNS lookup will failed and throw an error which can be seen in fidder response headers like this: "HTTP/1.1 502 Fiddler - DNS Lookup Failed". Under this situation I can not login to the game.
When I reconnect the Internet the responce headers will change into "HTTP/1.1 200 Connection Established", and the game can be logined.
I don't know if my guess is accurate, because I don't know much about network. Here is the file output by fidder, hope it works.
**Thanks to the developers who are concerned about this issue, I don't think it's very urgent and maybe in the future we can get a complete solution to this problem (or maybe it's just a glitch in my settings o(╥﹏╥)o).** 

[FidderPackage.zip](https://github.com/Grasscutters/Grasscutter/files/8598936/FidderPackage.zip)

lunaticwhat:
closed to clean up the workspace !

PR #845, opened 2022-05-12
Add block game update option

krrr:
## Description
Implemented fake update check API to block game update, enable the game to be played without internet connetion. current mitmproxy script (and fiddler script i guess) require dns lookup and will establish conneciton with official server (without sending data though). so it must be used together with my mitmproxy script 

this option is disabled by default as it needed to modify config file in JRE: https://www.petefreitag.com/item/916.cfm ,
and i have only tested on my up-to-date 2.6 global client

## Issues fixed by this PR
https://github.com/Grasscutters/Grasscutter/issues/326
https://github.com/Grasscutters/Grasscutter/issues/664

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
+++ b/src/main/java/emu/grasscutter/server/dispatch/DispatchServer.java
@@ -32,9 +32,11 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 
+import javax.net.ssl.SSLContext;
 import java.io.*;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.security.NoSuchAlgorithmException;
 import java.util.*;
 
 import static emu.grasscutter.utils.Language.translate;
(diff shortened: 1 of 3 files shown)


krrr:
my mitmproxy script:

```python
"""yuanshen mitm. need improve."""
from mitmproxy import http
from mitmproxy import ctx
from mitmproxy.proxy import layer, layers


def load(loader):
    ctx.options.web_open_browser = False
    # We change the connection strategy to lazy so that next_layer happens before we actually connect upstream.
    ctx.options.connection_strategy = "lazy"
    ctx.options.upstream_cert = False
    ctx.options.ssl_insecure = True

    
def next_layer(nextlayer: layer.NextLayer):
    ctx.log(
        f"{nextlayer.context=}\n"
        f"{nextlayer.data_client()[:70]=}\n"
    )
    sni = nextlayer.context.client.sni
    if nextlayer.context.client.tls and sni and (sni.endswith("yuanshen.com") or sni.endswith("mihoyo.com") or sni.endswith("hoyoverse.com")):
        ctx.log('sni:' + sni)
        nextlayer.context.server.address = ("127.0.0.1", 443)


def request(flow: http.HTTPFlow) -> None:
    # pretty_host takes the "Host" header of the request into account,
    # which is useful in transparent mode where we usually only have the IP
    # otherwise.
    if flow.request.pretty_url.startswith('http://log-upload-os.mihoyo.com'):
        flow.response = http.Response.make(
            404,  # (optional) status code
            b"404 not found",  # (optional) content
            {"Content-Type": "text/html"}  # (optional) headers
        )
        return

```

ghost:
All of this is completely unnecessary. We should remove `data\query_cur_region.txt` and just let server build one dynamically like it does right now when file is missing.

For a quick test, can you drop `query_cur_region.txt` from `data` directory, restart server and then start game without any of your modifications? I can not reproduce any attempt for connection to `autopatchhk.yuanshen.com` in this way since client is not given any source to download data from nor it is told what version server expects, so it just skips whole update process.

mcxiaoke:
grasscutter should block all requests to mhy official server, should rewrite autopatchhk.yuanshen.com urls to local server in query_cur_region

ghost:
There is no reason to provide them at all

krrr:
> All of this is completely unnecessary. We should remove `data\query_cur_region.txt` and just let server build one dynamically like it does right now when file is missing.
> 
> For a quick test, can you drop `query_cur_region.txt` from `data` directory, restart server and then start game without any of your modifications? I can not reproduce any attempt for connection to `autopatchhk.yuanshen.com` in this way since client is not given any source to download data from nor it is told what version server expects, so it just skips whole update process.

Your argument is complete nonsense. Should you ever tried yourself, you will easily find that the game refuse to login because of update failure after removing the file. error code 4205.

krrr:
> There is no reason to provide them at all

What's your problem?

krrr:
> grasscutter should block all requests to mhy official server, should rewrite autopatchhk.yuanshen.com urls to local server in query_cur_region

it shounld not be enabled by default. it requires modifying java runtime config file to enable TLSv1 when encryption enabled, otherwise login will fail.
Alternatively, let dispatch server bind other plain HTTP port when encryption enabled, and handle game update API using that port, to get around TLSv1 problem

ghost:
If you look at `RegionInfo.proto`, you will notice that only fields 1, 2 and 23 are necessary. Default `query_cur_region.txt` we provide is a copy from vanilla server and contains 1, 2, 3, 7, 8, 9, 10, 12, 14, 18, 19, 20, 22, 23, 24, 26, 27, 27, 31, 32, 33. We override fields 1, 2 and 23 however we leave rest of them as-is. 

Of these, 8 and 9 give client full URL to `autopatchhk.yuanshen.com`, while 12 give `2.6_live` portion. Fields 14 and 18 tell client to which version it should update itself while 19 and 20 give MD5s for them. I won't bother you with details of rest of fields, they contain additional URLs for stuff like payment portal or feedback.

Since we do not support any of this, we should not supply these fields and the client will ignore them. Similarly, not supplying update version and resources URL will have client skip update and just join server using whatever version it is at.

PR #861, opened 2022-05-13
enable script in big world

Akka0:
## Description

- support enable lua script in big world, both gadgets and monsters can be spawned (not enable default)

Notice: To enable it, first set config "enableScriptInBigWorld" to true

**since the pull, I have optimized the performance:**
now the group is loaded when a player near its position
cache the objects Lua created, it won't load again

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,6 +86,9 @@ dependencies {
 
     implementation group: 'org.luaj', name: 'luaj-jse', version: '3.0.1'
 
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.12.0'
+    implementation group:  'org.tinspin', name : 'tinspin-indexes', version: '1.7.1'
+
     protobuf files('proto/')
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -86,8 +86,7 @@ dependencies {
 
     implementation group: 'org.luaj', name: 'luaj-jse', version: '3.0.1'
 
-    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.12.0'
-    implementation group:  'org.tinspin', name : 'tinspin-indexes', version: '1.7.1'
+    implementation group:  'ch.ethz.globis.phtree', name : 'phtree', version: '2.5.0'
 
     protobuf files('proto/')
 }
--- a/src/main/java/emu/grasscutter/game/world/Scene.java
+++ b/src/main/java/emu/grasscutter/game/world/Scene.java
@@ -20,6 +20,7 @@ import emu.grasscutter.game.world.SpawnDataEntry.SpawnGroupEntry;
 import emu.grasscutter.net.packet.BasePacket;
 import emu.grasscutter.net.proto.AttackResultOuterClass.AttackResult;
 import emu.grasscutter.net.proto.VisionTypeOuterClass.VisionType;
+import emu.grasscutter.scripts.SceneIndexManager;
 import emu.grasscutter.scripts.SceneScriptManager;
 import emu.grasscutter.scripts.data.SceneBlock;
 import emu.grasscutter.scripts.data.SceneGroup;
@@ -491,7 +492,8 @@ public class Scene {
 		}
 	}
 	public Set<SceneBlock> getPlayerActiveBlocks(Player player){
-		return getScriptManager().getBlocks().stream()
+		// TODO consider the borders of blocks
+		return getScriptManager().getBlocks().values().stream()
 				.filter(block -> block.contains(player.getPos()))
 				.collect(Collectors.toSet());
 	}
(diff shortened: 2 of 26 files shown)


src/main/java/emu/grasscutter/utils/ConfigContainer.java
```diff
@@ -136,6 +136,8 @@ public static class Game {
         public int bindPort = 22102;
         /* This is the port used in the default region. */
         public int accessPort = 0;
+        public boolean enableScriptInBigWorld = false;
+        public int eventBusPoolSize = 8;
```
ghost: A thread pool is unnecessary for the main tick thread because ticks should never being happening at the same time.

src/main/java/emu/grasscutter/game/world/Scene.java
```diff
@@ -318,19 +326,28 @@ private void addEntityDirectly(GameEntity entity) {
...
+		if(entities == null || entities.isEmpty()){
+			return;
+		}
 		for (GameEntity entity : entities) {
 			this.addEntityDirectly(entity);
 		}
 		
-		this.broadcastPacket(new PacketSceneEntityAppearNotify(entities, VisionType.VISION_BORN));
+		this.broadcastPacket(new PacketSceneEntityAppearNotify(entities, visionType));
+		entities.forEach(scriptManager::callCreateEvent);
```
ghost: If you put callCreateEvent in Scene::addEntities, it would be called even when entites are spawned outside of the script handlers, which should not be happening. Instead, you should put it back where the ScriptManager spawns monsters/gadgets,

src/main/java/emu/grasscutter/game/world/Scene.java
```diff
@@ -484,49 +509,72 @@ public void checkSpawns() {
...
 		Set<SceneBlock> visible = new HashSet<>();
-		
 		for (Player player : this.getPlayers()) {
-			for (SceneBlock block : getScriptManager().getBlocks()) {
-				if (!block.contains(player.getPos())) {
-					continue;
+			var blocks = getPlayerActiveBlocks(player);
+			visible.addAll(blocks);
+
+			for(var block : blocks){
```
ghost: This loop for looping through blocks needs to be outside the player loop, otherwise you would be loading multiples of the same blocks if multiple people were in the same area.

ghost (changes requested):

src/main/java/emu/grasscutter/utils/ConfigContainer.java
Akka0: okay, it's reverted now

src/main/java/emu/grasscutter/game/world/Scene.java
Akka0: Moved to ScriptManager.addEntities

src/main/java/emu/grasscutter/utils/ConfigContainer.java
```diff
@@ -136,6 +136,7 @@ public static class Game {
         public int bindPort = 22102;
         /* This is the port used in the default region. */
         public int accessPort = 0;
+        public boolean enableScriptInBigWorld = false;
```
ghost, line 139: What's the point of making this controllable? I can't imagine situation where you wouldn't want this to be enabled, and we already depend on Lua for other stuff so...

src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
```diff
@@ -407,6 +396,9 @@ public void callEvent(int eventType, ScriptArgs params) {
 			}
 			
 			if (ret.isboolean() && ret.checkboolean()) {
+				if(StringUtils.isEmpty(trigger.action)){
```
ghost: Why are you using external library for such simple task? Java has native `.isEmpty()` for `String`

build.gradle
```diff
@@ -86,6 +86,9 @@ dependencies {
 
     implementation group: 'org.luaj', name: 'luaj-jse', version: '3.0.1'
 
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.12.0'
+    implementation group:  'org.tinspin', name : 'tinspin-indexes', version: '1.7.1'
```
ghost: This seems unused? Can't find any imports of `org.tinspin`.

ghost (changes requested):

src/main/java/emu/grasscutter/utils/ConfigContainer.java
Akka0: it's experimental and there are many Lua functions not implemented currently in the big world and it will produce lots of errors. 
So I make it disable default so people can open it if they really need it.

src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
Akka0: here I want to check if it is null or empty, now apache commons is removed

build.gradle
Akka0: I used the phtree inside the library to query the neighbor entities of player, now I extracted it from org.tinspin

mcxiaoke:
some one said this pr cause oom error with -Xmax12G, it has big problems.

Akka0:
since the pull, I have optimized the performance:
- now the group is loaded when a player near its position
- cache the objects Lua created, it won't load again

PR #865, opened 2022-05-13
Fix Expedition

XTL676:
## Description
*  make expeditionCountLimit change by player's adventure level(when player level up)
*  add avatar's expedition effect(such as -25% expedition time)
## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


data/ExpeditionAvatarEffect.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  {
+    "avatarId": 10000063,
+    "area": "Liyue",
+    "shortenRatio": 0,
+    "rewardRatio": 0.25
+  },
+  {
+    "avatarId": 10000056,
+    "area": "Inazuma",
```
ghost, line 34: Using game ID might be better and will be universally understood without having to rely on language-specific name mapping.

data/ExpeditionAvatarEffect.json
```diff
@@ -0,0 +1,38 @@
+[
```
ghost, line 1: Data for expedition avatar effect can be found in ProudSkillExcelConfigData, there is no need to hardcode avatar expedition effects one by one.

ghost (changes requested):

data/ExpeditionAvatarEffect.json
XTL676: Ah, yes, yes, yes, what you say is right. I really hope u or someone can fix the expedition,bye!

PR #884, opened 2022-05-14
Enable script in big world

Akka0:
## Description
since many changes and optimizations were introduced, I reopen a new one.

- support enable Lua script in the big world, both gadgets and monsters can be spawned (not enable default because there are many Lua funcs not impl currently)
- optimized the performance, including the groups & blocks are loaded when a player is near its position, and cache the objects Lua created, it won't load again (but the script will eval again to support ScriptLib trigger), use soft references to avoid OOM

Notice: To enable it, first set config "enableScriptInBigWorld" to true

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


ghost:
Thank you for your hard work with fixing the issues we outlined yesterday, There are some handlers and other core features I need to add. I will take a look tomorrow when I have more time and merge to a branch of development so we can work on it together.

ghost:
Ideally, the goal will be to have the scripts on permanently instead of being an option to toggle.

ghost:
>Ideally, the goal will be to have the scripts on permanently instead of being an option to toggle.

Need to make sure first we are not loading everything under the sun, lot of these are not supposed to be active unless specific conditions are met.

PR #892, opened 2022-05-14
Move gacha files to separate directory and refactor file serving

ghost:
## Description

- Move gacha related files to separate directory
- Do not cache template HTML, allow for on-the-fly editing and updating without restarting server
- Check if account exists before trying to query player in `gachaDetails`
- Send proper HTTP codes when account/player/template not found
- Fix broken argument order for `getGachaRecords`
- Properly handle files as UTF-8
- Fix language mapping (`record` was completely broken since it was using `en` instead of `en-us`, while `details` was using improper casing [we seem to save locale as `en-US` in database but generated mappings are casted to lowercase])

## Issues fixed by this PR

N/A

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/data/gacha/details.html
+++ b/data/gacha/details.html
@@ -84,7 +84,7 @@
             var fiveStarItems = {{FIVE_STARS}};
             var fourStarItems = {{FOUR_STARS}};
             var threeStarItems = {{THREE_STARS}};
-            var lang = "{{LANGUAGE}}";
+            var lang = "{{LANGUAGE}}".toLowerCase();
 
             function getNameForId(itemId) {
                 if (mappings[lang] != null && mappings[lang][itemId] != null) {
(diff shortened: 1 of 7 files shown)


ghost (approved):

mingjun97:
>Fix language mapping (record was completely broken since it was using en instead of en-us, while details was using improper casing [we seem to save locale as en-US in database but generated mappings are casted to lowercase])

The `en-us` is given by certain anime game client actually. The when the in-game webview try to access web pages, it will give GET parameters like `lang=[ISO639 language code]`. The original implement was tyring to let the user decide which language they want to see.

ghost:
>The en-us is given by certain anime game client actually. The when the in-game webview try to access web pages, it will give GET parameters like lang=[ISO639 language code].

Was simply `en` for me :>

PR #893, opened 2022-05-14
Remove the extra slash while loading resources.

line-bear:
## Description
The new default config file already contains / at the end of path string, which may cause double slash while loading resources, It seems to work fine at the moment, but it is recommended to remove it to prevent unpredictable situations.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


ghost:
We can not be sure that someone editing config file will remember to add `/` at the end of string.

src/main/java/emu/grasscutter/Configuration.java
```diff
@@ -48,15 +48,15 @@ public final class Configuration extends ConfigContainer {
      */
     
     public static String DATA(String path) {
-        return DATA_FOLDER + "/" + path;
+        return DATA_FOLDER + path;
```
ghost:
```suggestion
        return Paths.get(DATA_FOLDER, path).toString();
```

src/main/java/emu/grasscutter/Configuration.java
```diff
@@ -48,15 +48,15 @@ public final class Configuration extends ConfigContainer {
...
     }
     
     public static String RESOURCE(String path) {
-        return RESOURCES_FOLDER + "/" + path;
+        return RESOURCES_FOLDER + path;
```
ghost:
```suggestion
        return Paths.get(RESOURCES_FOLDER, path).toString();
```

src/main/java/emu/grasscutter/Configuration.java
```diff
@@ -48,15 +48,15 @@ public final class Configuration extends ConfigContainer {
...
     }
     
     public static String SCRIPT(String path) {
-        return SCRIPTS_FOLDER + "/" + path;
+        return SCRIPTS_FOLDER + path;
```
ghost:
```suggestion
        return Paths.get(SCRIPTS_FOLDER, path).toString();
```

ghost (changes requested):
Let's use proper path joining.

```java
import java.nio.file.Paths;
```

ghost:
You should also perform same changes in all other files where `*_FOLDER` are used. Most likely we should mark them as private and instead provide helper functions like `DATA` for all other ones that don't have one currently.

line-bear (commented):
Use path join to prevent double slash

line-bear:
> You should also perform same changes in all other files where `*_FOLDER` are used. Most likely we should mark them as private and instead provide helper functions like `DATA` for all other ones that don't have one currently.

all done, have fun

line-bear (commented):
Mark fields as private to prevent use *FOLDER directly

src/main/java/emu/grasscutter/utils/Crypto.java
```diff
@@ -20,11 +20,11 @@ public final class Crypto {
 	public static byte[] ENCRYPT_SEED_BUFFER = new byte[0];
 	
 	public static void loadKeys() {
-		DISPATCH_KEY = FileUtils.read(KEYS_FOLDER + "/dispatchKey.bin");
-		DISPATCH_SEED = FileUtils.read(KEYS_FOLDER + "/dispatchSeed.bin");
+		DISPATCH_KEY = FileUtils.read(KEY("/dispatchKey.bin"));
```
ghost: No need for slash here

ghost:
Thanks. Looks good to me now.

I'll most likely go around whole codebase in another PR later and see if we can refactor `Utils.toFilePath` to be a wrapper for path join instead of dumb replacing `/` with separator. I've also spotted few places that forget to use `Utils.toFilePath` so will need to handle that too.

line-bear:
> Thanks. Looks good to me now.
> 
> I'll most likely go around whole codebase in another PR later and see if we can refactor `Utils.toFilePath` to be a wrapper for path join instead of dumb replacing `/` with separator. I've also spotted few places that forget to use `Utils.toFilePath` so will need to handle that too.

Waitting for your message :)

ghost (approved):

Issue #919, opened 2022-05-15
[Bug] Can not change default IP address after upgrade to newest dev branch.


nigpo:
After upgrade to newest branch and change the server ip address that I wanna, I will lost connection of game after login.
We can change GameServer java file and HttpServer and remove host for bug fixed.

ghost:
Have you checked your config file? We did a config file rework a while ago.

ghost:
#993

nigpo:
Solve with latest dev version.

nigpo:
#993 Solved.

PR #921, opened 2022-05-15
Fix connection bugs

nigpo:
Connection will lost after add host for default, local device and remote device which that not connect by host will unable to access game.


nigpo:
#919

KingRainbow44 (changes requested):
please don't remove the ability to bind to a specific address. this is an intended feature.

ghost:
We will not be removing the ability for a server to bind to an IP.

PR #929, opened 2022-05-15
Enhance the default permissions logic

omg-xtao:
## Description

Make default permissions not write to the database and be able to alter them dynamically.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/auth/DefaultAuthenticators.java
+++ b/src/main/java/emu/grasscutter/auth/DefaultAuthenticators.java
@@ -41,10 +41,6 @@ public final class DefaultAuthenticators {
                     responseMessage = translate("messages.dispatch.account.username_create_error");
                     Grasscutter.getLogger().info(translate("messages.dispatch.account.account_login_create_error", address));
                 } else {
-                    // Add default permissions.
-                    for (var permission : ACCOUNT.defaultPermissions)
-                        account.addPermission(permission);
-
                     // Continue with login.
                     successfulLogin = true;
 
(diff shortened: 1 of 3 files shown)


ghost:
This change makes more sense but I'd like opinions from other people about this before we merge.

jie65535:
I think many GM need this because the current situation is that it is very troublesome to change the default permissions of old users. For example, a new plug-in has been added with new permissions, which is a nightmare.

Birdulon:
> This change makes more sense but I'd like opinions from other people about this before we merge.

Perhaps it would be useful to go one step further and have permission group aliases so that admins can define permission groups like `user`, `poweruser`, `gm`, and then add whatever permissions their plugins use to those groups as they please.

exzork:
oh god, if this permission is updated it's mean we have to update the database :(

omg-xtao:
> oh god, if this permission is updated it's mean we have to update the database :(

Why? Both new and existing users can have the default permissions.

exzork:
For the default one sure it's, but for custom permission i think no.

Birdulon:
> oh god, if this permission is updated it's mean we have to update the database :(

Under the current system yes, under this PR's system no.

Current system: 
 - User is created, default permissions are added to their account and saved in the db for later.
 - When default permissions are changed, admin has to either:
   - leave old users without these changes, or
   - go through and update all users' individual permissions in the db.

This system:
 - User is created with no attached permissions.
 - Permission checks are performed against not only the individual user, but also against the defaults.
 - When default permissions are changed, they apply to all future permission checks for all users without any db updates.

ghost (approved):

PR #979, opened 2022-05-18
The show command was added to facilitate comparison of ID information

cfuncode:
The show command was added to facilitate comparison of ID information

有很多ID还没有对应上名称, 不过不影响使用

There are many ids that don't have names, but it doesn't affect use

https://1drv.ms/v/s!AhrE0Sk9D3ZprHMTXyWftwJFS5Al?e=XNFdUY


cfuncode:
还有很多名称信息没翻译,太繁琐了.

cfuncode:
> 还有很多名称信息没翻译,太繁琐了.

翻译好了。pl-PL语言没有翻译

cfuncode:
en-US/pl-PL 的 Talents 没有对照，懒得一个个 OCR 了，太累了。

Tesutarin:
你的 `en-US.json` 和 `zh-CN.json` 已过时，需要更新
`en-US.json` and `zh-CN.json` are outdated, need to be updated

cfuncode:
> 你的 `en-US.json` 和 `zh-CN.json` 已过时，需要更新 `en-US.json` and `zh-CN.json` are outdated, need to be updated

Alright

arub3do:
Perhaps the data files in `showinfos/*` could be generated dynamically, or cached when GC is first opened.
Then we wouldn't need to update those files every time.

cfuncode:
> Perhaps the data files in `showinfos/*` could be generated dynamically, or cached when GC is first opened. Then we wouldn't need to update those files every time.

No, this is not convenient for internationalization, and many dynamically generated information names are empty, and also not conducive to the subsequent version of the update to expand.

cfuncode:
[You can use the show command to make it even simpler](https://github.com/Grasscutters/Grasscutter/pull/979/commits/9a6e2717ed043f417cb46c31a920cc7c733bc5a3)
For example:

```sh
/show art 1 10 flower
/show art 1 flower
/show art flower
```

The effect is the same.

KingRainbow44:
will be closing this PR since it is inactive. if there are any updates please re-open this PR or create another one.

Issue #1005, opened 2022-05-20
[错误] Android client login interface becomes code format


izhuolin:
The Android client login interface has changed to code format, which has been the same since yesterday's version.
![75E43246-347A-41FF-9BE2-C94FD70F47B8](https://user-images.githubusercontent.com/61179498/169430197-35729fa2-5e62-4788-bb61-d205162757d5.jpeg)
![EC9F5989-8A73-49A0-983D-5406CC73FB76](https://user-images.githubusercontent.com/61179498/169430259-15840930-490f-4acd-ae48-708e1e196637.jpeg)

PR #1010, opened 2022-05-20
fix version.json

2y8e9h22:
login er

## Description

## Type of changes

- [ √] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] My pull request is unique and no other pull requests have been opened for these changes
- [ ] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [ ] I am responsible for any copyright issues with my code if it occurs in the future.

### DIFF
--- a/src/main/java/emu/grasscutter/server/http/handlers/GenericHandler.java
+++ b/src/main/java/emu/grasscutter/server/http/handlers/GenericHandler.java
@@ -41,7 +41,7 @@ public final class GenericHandler implements Router {
         express.all("/perf/config/verify", new HttpJsonResponse("{\"code\":0}"));
 
         // webstatic-sea.hoyoverse.com
-        express.get("/admin/mi18n/plat_oversea/*", new HttpJsonResponse("{\"version\":51}"));
+        express.get("/admin/mi18n/plat_oversea/m202003048/m202003048-version.json", new HttpJsonResponse("{\"version\":51}"));
         
         express.get("/status/server", GenericHandler::serverStatus);
     }


exzork:
Can you take a look and test this? @4Benj

src/main/java/emu/grasscutter/server/http/handlers/GenericHandler.java
```diff
@@ -41,7 +41,7 @@ public final class GenericHandler implements Router {
         express.all("/perf/config/verify", new HttpJsonResponse("{\"code\":0}"));
 
         // webstatic-sea.hoyoverse.com
-        express.get("/admin/mi18n/plat_oversea/*", new HttpJsonResponse("{\"version\":51}"));
+        express.get("/admin/mi18n/plat_oversea/m202003048/m202003048-version.json", new HttpJsonResponse("{\"version\":51}"));
```
4Benj, line 44: The reason this is a wildcard is because there are multiple versions of .json, typically it differs per platform but I know there are some some other difference but I don't recall on how they occur.
What exactly is the issue with the current wildcard solution? (other then there are other files (-en.json) that need to be accessed, but I will be coming up with a fix for that soon.)

exzork: #1005 i think

4Benj: > #1005 i think

Yeah I think that is related to the thing I mentioned about -en.json and related files.
When I made the wildcard fix before I didn't realise that there were translation files within there as well.
I already have a fix in mind for this without removing the wildcard and causing a bunch of other issues. Hopefully I will have it done later today.

4Benj: This is the type of file I didn't realise were needed
https://webstatic-sea.hoyoverse.com/admin/mi18n/plat_oversea/m202003048/m202003048-en.json

mcxiaoke:
ref https://github.com/Grasscutters/Grasscutter/issues/982

4Benj:
The purposed fix will only solve things temporarily and will cause errors on some clients. I will submit a new PR later today with a better solution that won’t cause errors on other clients.

PR #1022, opened 2022-05-21
Allow commands to target offline players

Birdulon:
## Description
Some commands do not strictly require the target player to be online in order to function. Some don't need a target player at all. Others in the future may want to only target offline players. This PR adds a new annotation to commands such that these different requirements can be expressed more easily. It should also aid with future overhauls like #751 that will want this information ahead of time, rather than just as an error when the commands are called with incorrect targets.

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


ghost (approved):

Birdulon:
Sets `/account`, `/broadcast`, `/help`, `/language`, `/list`, `/reload`, `/restart`, `/sendmail` and `/stop` to `Command.TargetRequirement.NONE`, everything else is using the default value of `Command.TargetRequirement.ONLINE` for now. Changes to offline `Player` loading will need to be made before `/permission` and other commands that make sense to use on offline players can be set to `Command.TargetRequirement.PLAYER`.

PR #1027, opened 2022-05-21
add support for codex animal, weapon, reliquary

ShigemoriHakura:
## Description

add support for codex animal, weapon, reliquary
fix the error when running in idea
fix the translation error in givechar

## Type of changes

- [x] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


arub3do:
I guess these `CmdId` is based on V2.2.0?
```
Proto.QueryCodexMonsterBeKilledNumReq,4201,
Proto.QueryCodexMonsterBeKilledNumRsp,4210,
```

Maybe V2.6 should be: 4207/4208

ShigemoriHakura:
> I guess these `CmdId` is based on V2.2.0?
> 
> ```
> Proto.QueryCodexMonsterBeKilledNumReq,4201,
> Proto.QueryCodexMonsterBeKilledNumRsp,4210,
> ```
> 
> Maybe V2.6 should be: 4207/4208

The problem is that only the first monster in QueryCodexMonsterBeKilledNumReq's codexIdList can show the correct beKilledNum. I think it's not due to the cmdid mismatch.

src/main/java/emu/grasscutter/utils/Utils.java
```diff
@@ -200,8 +201,10 @@ public static void startupCheck() {
 			createFolder(dataFolder);
 
 		// Make sure the data folder is populated, if there are any missing files copy them from resources
-		DataLoader.CheckAllFiles();
-
+		if(Grasscutter.class.getResource("").getProtocol() == "jar"){
```
ghost, line 204: Could you explain this change?

ghost (changes requested):

src/main/java/emu/grasscutter/utils/Utils.java
ShigemoriHakura: DataLoader.CheckAllFiles() only works when running in jar (grasscutter-xxx-dev.jar),

Otherwise it will throw out an error:
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because "filenames" is null
	at emu.grasscutter.data.DataLoader.CheckAllFiles(DataLoader.java:58)
	at emu.grasscutter.utils.Utils.startupCheck(Utils.java:208)
	at emu.grasscutter.Grasscutter.<clinit>(Grasscutter.java:77)

Can be reproduced by using .\gradlew.bat run

ghost: I think this needs more attention and be fixed properly in separate PR. Your fix will just ensure it won't crash but it won't copy necessary files. Also the log message is pointless.

ShigemoriHakura: I know, but I think those who uses ide to run grasscutter know how to copy the data from resources to /data

TalexDreamSoul:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

ShigemoriHakura:
Refactoring

Issue #994, opened 2022-05-19
Main characters elemental bursts doesn't work.


Rallus-Zenturi:
Main characters Aether ; Lumine ANY elemental burst doesn't work. 
Though it was embryo issue but no, if you'll delete burst ability in other character, the elemental burst icon still will glow and can be initiated. It might be some kind of energy issue.

Branch used - Grasscutter version: 1.1.2-dev

**Screenshots**
![image](https://user-images.githubusercontent.com/43311503/169302360-6385c39b-6f92-477a-9c06-5ae9b005099c.png)
![image](https://user-images.githubusercontent.com/43311503/169302384-ba166c5f-f6fb-441c-b531-4ac5ad139c8e.png)
![image](https://user-images.githubusercontent.com/43311503/169303485-83f684be-7fa2-48eb-9a78-e87385549ef3.png)

Additionally I would like to suggest implementation of feature
Interacting with Statues of the Seven and allowing to players to switch main characters skilldepots without editing db manually.

Birdulon:
Traveler can't gain any energy even when you set their fightprops, it might be related to quest flags on the client.

lunaticwhat:
all are known issues, thanks.

PR #1029, opened 2022-05-21
Implement Some Energy-related Stuff

GanyusLeftHorn:
## Description

This PR implements and fixes some energy-related things. In particular:
* Energy particles created by elemental skills now are of the correct elemental type (were elementless before).
* Picking up energy particles or orbs now gives the correct amount of energy depending on character element, and also now honors the character's energy recharge stat.
* Using an elemental burst now depletes energy. Since by far not all energy-related features are implemented yet, this is configurable via a new `server.game.game.gameOptions.energyUsage` settings in `config.json`, which default to `false`, preserving the current behavior of bursts always being full as the default.
* Fixes the traveller not being able to gain energy (issue #994).

Known issues and energy stuff that is still missing (not necessarily an exhaustive list):
* Some character skills don't generate the correct amount of particles. As far as I can tell, this is caused by `ClientAbilityInitFinishNotify` not having the correct number of `ABILITY_ACTION_GENERATE_ELEM_BALL` entries. I don't know yet what is causing this to be the case for some, but not all, skills.
* Some energy generation by weapon passives, like that of the Favonius weapons, does not work correctly.
* We don't generate energy for NAs/CAs yet.
* Enemies don't drop energy particles/orbs yet.

## Issues fixed by this PR

#994

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


ghost (approved):

src/main/java/emu/grasscutter/game/managers/EnergyManager.java
```diff
@@ -0,0 +1,215 @@
...
+		for (int i = 0; i < this.player.getTeamManager().getActiveTeam().size(); i++) {
+			EntityAvatar entity = this.player.getTeamManager().getActiveTeam().get(i);
+
+			// On-field vs off-field multiplier.
+			// The on-field character gets no penalty.
+			// Off-field characters get a penalty depending on the team size, as follows:
+			//	  - 4 character team: 0.6
+			//	  - 3 character team: 0.7
+			//    - 2 character team: 0.8
+			float offFieldPenalty = (this.player.getTeamManager().getCurrentCharacterIndex() == i) ? 1.0f : 1.0f - this.player.getTeamManager().getActiveTeam().size() * 0.1f;
```
ghost: Make sure `offFieldPenalty` does not go below 0 (or .1) in case someone has more than 4 avatars in their party.

src/main/java/emu/grasscutter/game/managers/EnergyManager.java
```diff
@@ -0,0 +1,215 @@
...
+		// Determine the element of the energy particle that we have to generate.
+		// In case we can't, we default to an elementless particle.
+		// The element is the element of the avatar that has cast the ability.
+		// We can get that from the avatar's skill depot.
+		int itemId = 2024;
+		
+		// Try to fetch the avatar from the player's party and determine their element.
+		// ToDo: Does this work in co-op?
+		int avatarId = getCastingAvatarIdForElemBall(invoke.getEntityId());	
+		Optional<EntityAvatar> avatarEntity = player.getTeamManager().getActiveTeam()
```
ghost, line 88: It is reccomended to use `TeamManager::getCurrentAvatarEntity()`

ghost (changes requested):

src/main/java/emu/grasscutter/game/managers/EnergyManager.java
GanyusLeftHorn: Oh, good point, I didn't consider that. Do you happen to know what the official server does when we have a 5th trial character in the party? Does it continue to scale down, or does it continue doing what it does for a 4 character party?

src/main/java/emu/grasscutter/game/managers/EnergyManager.java
GanyusLeftHorn: But that only gives me the currently active avatar, doesn't it? The character generating the particle doesn't have to be active (think for example Zhongli's pillar generating a particle while he is off field).

ghost (approved):

PR #1033, opened 2022-05-22
fix disconnection when udp client port changed

zhaodice:
related : https://github.com/Melledy/kcp-netty/pull/1
if you don't trust my compiling, please look at above pull request and do self-compiling


zhaodice:
I have tested it that it would resolve disconnection I encountered

ghost:
Will make a commit myself, thank you for your contribution.

PR #1101, opened 2022-05-30
Made Energy Generation for Kuki and Yelan more accurate

honeyed-lemons:
## Description
 - Updated energy generation to be more accurate to calculations from Ubatcha
## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

- [x]  My code follows the style guidelines of this project

### DIFF
--- a/src/main/resources/defaults/data/SkillParticleGeneration.json
+++ b/src/main/resources/defaults/data/SkillParticleGeneration.json
@@ -515,7 +515,11 @@
 		"amountList": [
 			{
 				"value": 4,
-				"chance": 100
+				"chance": 570
+			},
+			{
+				"value": 3,
+				"chance": 30
 			}
 		]
 	},
@@ -555,7 +559,7 @@
 		"amountList": [
 			{
 				"value": 1,
-				"chance": 100
+				"chance": 45
 			}
 		]
 	},
--- a/src/main/resources/defaults/data/SkillParticleGeneration.json
+++ b/src/main/resources/defaults/data/SkillParticleGeneration.json
@@ -515,7 +515,7 @@
 		"amountList": [
 			{
 				"value": 4,
-				"chance": 570
+				"chance": 70
 			},
 			{
 				"value": 3,


src/main/resources/defaults/data/SkillParticleGeneration.json
```diff
@@ -555,7 +559,7 @@
 		"amountList": [
 			{
 				"value": 1,
-				"chance": 100
+				"chance": 45
```
GanyusLeftHorn, line 562: The zero particle case is handled by the client. Shinobu always generates exactly one particle when the client invokes GENERATE_ELEM_BALL.

src/main/resources/defaults/data/SkillParticleGeneration.json
```diff
@@ -515,7 +515,11 @@
 		"amountList": [
 			{
 				"value": 4,
-				"chance": 100
+				"chance": 70
+			},
+			{
+				"value": 3,
+				"chance": 30
```
GanyusLeftHorn, line 522: Do you have a source for this? I know these are the rates that some leaks mentioned, but `BinOutput/Ability/Temp/AvatarAbilities/ConfigAbility_Avatar_Yelan.json` does not support this. The data in there says that she always generates 12 energy, i.e. 4 hydro particles.

src/main/resources/defaults/data/SkillParticleGeneration.json
honeyed-lemons: Ah, ok, got it

src/main/resources/defaults/data/SkillParticleGeneration.json
honeyed-lemons: Theres always a chance it changed from beta, however ubacha has claimed that theres a chance for it to generate 3-4 depending on hold and press skill. Heres the tweet. Unless his source lied it'd be weird if it generated anything else https://twitter.com/Ubatcha1/status/1509954923357777921?s=20&t=x5TRifdlh_japR82Ju4Oyg

GanyusLeftHorn: Yes, that is the leak I mentioned. There was also an earlier one about her always generating four, from an earlier version of the beta. It is possible that they just went back to that for the release.

For now, I would suggest sticking to what the data in `BinOutput` says. If tests on the live server show different behavior, we can always change it then.

honeyed-lemons: 👍 got it

PR #1137, opened 2022-06-01
Fix deforestation protocol and add a new field

zhaodice:
before I have analyzed a wrong pack, this is right protocol file,and,
I don't understand the reason why ** makes two same position fields


GanyusLeftHorn:
One is the tree position, one the drop position.

ghost (approved):

zhaodice:
> One is the tree position, one the drop position.

oh,but the data is same,I can't understand

PR #1164, opened 2022-06-02
Move some data into the jar.

GanyusLeftHorn:
## Description

Right now, `data` is in a strange place. It contains a mixture of files that users should be able to edit on one hand, and files that are integral to the workings of the server on the other hand. Frequently, new features and/or bug fixes rely on changes to files that currently reside in `data`.

However, files in `data` can not only be edited by the user, they are also never updated once they are created. Our startup check only makes sure that the files exists, and copies them from within the jar to the `data` folder if they don't. This can create a situation where a user has a build in which a particular bug has already been fixed, or a particular feature has already been added, but they are still experiencing the bug or missing the feature because their files in `data` are outdated.

For this reason, I would propose that many of the files currently residing in `data` should be moved to within the jar, so that they get updated with every build. Of course, this comes at the cost of users not being able to edit these files so easily anymore. However, users who are sufficiently advanced to be able to correctly edit these files can probably also be expected to rebuild the server if they want to customize them.

This PR moves `EnergyDrop.json` and `SkillParticleGeneration.json` into the jar, since I know exactly how those are used. However, many of the other files in `data` would probably also benefit from being moved (thinking of things like `Drop.json`, `ExpeditionRewards.json`, `Shop.json` etc.). If desired, I'd be happy to move further files and make the necessary changes to the places in the code where they are used. Just let me know and I will add it to this PR.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


WhisperWithDelight:
Maybe it is better to check if there is a corresponding .json file in the data folder first, and if so we use the .json configuration in the data folder, if not we use the built-in configuration in .jar as you proposed (instead of copying the .json file into the data folder as we are doing now). So that for people who don't want to customize the configuration, their data folder is empty and can update the configuration with every build, while for people who want to  customize the configuration, they can easily add the .json configuration they want.

ghost:
Not a big fan of this, we should let people edit the data in their servers as easily and as much as possible (From excels to data files).

One thing i do dislike however, is how people need to manually copy their data files over if there was an update to them.

GanyusLeftHorn:
Okay, should I close this or do you think what @WhisperWithDelight proposed would be a viable alternative?

ghost:
Lets close it for now, I want to discuss this with you and @4Benj later.

PR #1181, opened 2022-06-03
List game & server version when starting server

AZthemute:
## Description

## Issues fixed by this PR
Outputs game and server version on startup. Currently hardcoded in the language files unfortunately. If anyone wants to fix my idiocy, I'd be pleased to accept it.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/server/game/GameServer.java
+++ b/src/main/java/emu/grasscutter/server/game/GameServer.java
@@ -270,8 +270,6 @@ public final class GameServer extends KcpServer {
 	@Override
 	public void onStartFinish() {
 		Grasscutter.getLogger().info(translate("messages.status.free_software"));
-		Grasscutter.getLogger().info(translate("messages.status.game_version", GameConstants.VERSION));
-		Grasscutter.getLogger().info(translate("messages.status.version", BuildConfig.VERSION, BuildConfig.GIT_HASH));
 		Grasscutter.getLogger().info(translate("messages.game.port_bind", Integer.toString(address.getPort())));
 		ServerStartEvent event = new ServerStartEvent(ServerEvent.Type.GAME, OffsetDateTime.now()); event.call();
 	}
--- a/src/main/java/emu/grasscutter/server/game/GameServer.java
+++ b/src/main/java/emu/grasscutter/server/game/GameServer.java
@@ -27,6 +27,7 @@ import emu.grasscutter.server.event.game.ServerTickEvent;
 import emu.grasscutter.server.event.internal.ServerStartEvent;
 import emu.grasscutter.server.event.internal.ServerStopEvent;
 import emu.grasscutter.task.TaskMap;
+import emu.grasscutter.BuildConfig;
 
 import java.net.InetSocketAddress;
 import java.time.OffsetDateTime;
@@ -269,8 +270,8 @@ public final class GameServer extends KcpServer {
 	@Override
 	public void onStartFinish() {
 		Grasscutter.getLogger().info(translate("messages.status.free_software"));
-		Grasscutter.getLogger().info(translate("messages.status.game_version"));
-		Grasscutter.getLogger().info(translate("messages.status.version"));
+		Grasscutter.getLogger().info(translate("messages.status.game_version", GameConstants.VERSION));
+		Grasscutter.getLogger().info(translate("messages.status.version", BuildConfig.VERSION));
 		Grasscutter.getLogger().info(translate("messages.game.port_bind", Integer.toString(address.getPort())));
 		ServerStartEvent event = new ServerStartEvent(ServerEvent.Type.GAME, OffsetDateTime.now()); event.call();
 	}
--- a/src/main/java/emu/grasscutter/server/game/GameServer.java
+++ b/src/main/java/emu/grasscutter/server/game/GameServer.java
@@ -271,7 +271,7 @@ public final class GameServer extends KcpServer {
 	public void onStartFinish() {
 		Grasscutter.getLogger().info(translate("messages.status.free_software"));
 		Grasscutter.getLogger().info(translate("messages.status.game_version", GameConstants.VERSION));
-		Grasscutter.getLogger().info(translate("messages.status.version", BuildConfig.VERSION));
+		Grasscutter.getLogger().info(translate("messages.status.version", BuildConfig.VERSION, BuildConfig.GIT_HASH));
 		Grasscutter.getLogger().info(translate("messages.game.port_bind", Integer.toString(address.getPort())));
 		ServerStartEvent event = new ServerStartEvent(ServerEvent.Type.GAME, OffsetDateTime.now()); event.call();
 	}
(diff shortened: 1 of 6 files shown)


src/main/resources/languages/en-US.json
```diff
@@ -52,7 +52,8 @@
       "run_mode_help": "Server run mode must be 'HYBRID', 'DISPATCH_ONLY', or 'GAME_ONLY'. Unable to start Grasscutter...",
       "create_resources": "Creating resources folder...",
       "resources_error": "Place a copy of 'BinOutput' and 'ExcelBinOutput' in the resources folder.",
-      "version": "Grasscutter version: %s-%s"
```
4Benj, line 55: Don't change this

src/main/java/emu/grasscutter/server/game/GameServer.java
```diff
@@ -269,6 +269,8 @@ public void run() {
 	@Override
 	public void onStartFinish() {
 		Grasscutter.getLogger().info(translate("messages.status.free_software"));
+		Grasscutter.getLogger().info(translate("messages.status.game_version"));
+		Grasscutter.getLogger().info(translate("messages.status.version"));
```
4Benj: Do something like this but edit it to use the new translation system. 
https://github.com/mingjun97/Grasscutter/blob/5aefad61eacfad17c46a4f139b5d5c8f56c0aff7/src/main/java/emu/grasscutter/Grasscutter.java#L84
(Originally from an old PR which was merged but was accidently removed)

4Benj (changes requested):

src/main/java/emu/grasscutter/server/game/GameServer.java
AZthemute: Thanks for the help, I've fiddled around with the code a little more and everything works perfectly on my side!

src/main/resources/languages/en-US.json
AZthemute: What's wrong?

src/main/resources/languages/zh-CN.json
```diff
@@ -52,7 +52,8 @@
       "run_mode_help": "服务器运行模式必须为 'HYBRID'(混合)、'DISPATCH_ONLY'(仅 Dispatch) 或 'GAME_ONLY'(仅游戏)。Grasscutter 启动失败...",
       "create_resources": "正在创建 resources 目录...",
       "resources_error": "请将 BinOutput 和 ExcelBinOutput 复制到 resources 目录。",
-      "version": "Grasscutter 版本：%s-%s"
+      "version": "Grasscutter 版本：%s",
+      "game_version": "游戏版本: 2.7.0"
```
Tesutarin: incorrect punctuation
`"游戏版本：%s"`
update other language files too

src/main/resources/languages/en-US.json
Birdulon: Changing the substitution parameter count would normally be a problem but I'm not getting any hits for existing usage of this so it might be fine.

Birdulon: I've traced it back to https://github.com/Grasscutters/Grasscutter/commit/edc75b2632318c99671cb089d6ed694750864c0c but it looks like the code that was referencing this translation string has since been replaced with https://github.com/mingjun97/Grasscutter/blob/2dc0120c81ff0c47af3f1e30069027d0622ea5c5/src/main/java/emu/grasscutter/Grasscutter.java#L84 so it's probably fine to repurpose this so long as the new version string is appropriate (I haven't checked that)

AZthemute: The new version string is `1.1.2-dev` so it wouldn't even be correct to use %s-%s.

src/main/resources/languages/zh-CN.json
AZthemute: Sorry about that, I just used a translator because I didn't want to leave the other languages throwing errors constantly.

src/main/resources/languages/en-US.json
4Benj: > The new version string is `1.1.2-dev` so it wouldn't even be correct to use %s-%s.

The second %s is for the git hash so we can easily identify someone’s build (primarily for development but also useful in stable)

AZthemute: > > The new version string is `1.1.2-dev` so it wouldn't even be correct to use %s-%s.
> 
> The second %s is for the git hash so we can easily identify someone’s build (primarily for development but also useful in stable)

Ohhhh sorry about that, I'll make sure to re-implement that as soon as I can.

AZthemute: Yup, it works for me now.

Birdulon: > > The new version string is `1.1.2-dev` so it wouldn't even be correct to use %s-%s.
> 
> The second %s is for the git hash so we can easily identify someone’s build (primarily for development but also useful in stable)

AFAICT it's only used to construct the version string, we may as well just have a single string for version and change the gradle script from `git rev-parse --verify --short HEAD` to `git describe --tags --dirty`

src/main/resources/languages/pl-PL.json
```diff
@@ -48,7 +48,8 @@
       "run_mode_help": "Tryb pracy serwera musi być ustawiony na 'HYBRID', 'DISPATCH_ONLY', lub 'GAME_ONLY'. Nie można wystartować Grasscutter...",
       "create_resources": "Tworzenie folderu resources...",
       "resources_error": "Umieść kopię 'BinOutput' i 'ExcelBinOutput' w folderze resources.",
-      "version": "Grasscutter versión: %s-%s"
+      "version": "Grasscutter versión: %s-%s",
+      "game_version": "Game versión: 2.7.0"
```
ghost: Please change these so the version number is a variable.

src/main/java/emu/grasscutter/server/game/GameServer.java
```diff
@@ -269,6 +270,8 @@ public void run() {
 	@Override
 	public void onStartFinish() {
 		Grasscutter.getLogger().info(translate("messages.status.free_software"));
+		Grasscutter.getLogger().info(translate("messages.status.game_version", GameConstants.VERSION));
+		Grasscutter.getLogger().info(translate("messages.status.version", BuildConfig.VERSION, BuildConfig.GIT_HASH));
```
ghost: The BuildConfig version log message should be run when the server starts not after it finishes starting up, in case the server throws an error before startup.

ghost (changes requested):

src/main/resources/languages/pl-PL.json
AZthemute: Done!

src/main/java/emu/grasscutter/server/game/GameServer.java
AZthemute: Done!

ghost (approved):

PR #1193, opened 2022-06-04
[WIP - not working properly yet] Implement expire time option in /sendmail command.

AZthemute:
## Description

## Issues fixed by this PR
/sendmail now lets you set an expire time... except it doesn't work.
The /sendmail command works, however sending the actual mail seems to break everything.

### Current observations:
*In Mail.java line 31, the expire time is currently hardcoded (in Grasscutters:development) to be 604800 seconds - 7 days. 
`this(new MailContent(), new ArrayList<MailItem>(), (int) Instant.now().getEpochSecond() + 604800);`
When I change this value, every mail has it's expiry time set to that value.
In my drafted changes, I have:
`this(new MailContent(), new ArrayList<MailItem>(), (int) Instant.now().getEpochSecond() + expireTime);`
...but this doesn't work, as sending mails with the command that has an expiry time of 604800 seconds prevents it from appearing in the mailbox. But weirdly enough, 7 days... kind of works. Which leads to my next observation...

*Only "days" will send the mail. However even then, the mail expires in 19147 days instead of what you input. This... really confuses me considering all that changes with the time format (seconds, minutes, etc.) is how much the number the user inputted is multiplied by.

**Going on from this point, any other input other than "days" will send the mail notification and put the red dot on the mail icon, but there will be nothing in the mail tab (the red dot is not cleared either after opening the mail tab). This also occurs when you set a mail to expire in 19147 or more days. Also no, the mail is not saved in the database. It's possible that it's being deleted instantly.

*It's even worse than that, though. 7777777 seconds sends the mail with an expiry time of 7732 days and trying my 7 days example after that gives the next mail I send an expiry time of 26879 days, proving that **the expiry times add up somewhere. I have no clue where.**

*When I relog, the expiry times LOOK to all be the same. However, they have the correct expireTime in the database. This is also true BEFORE relogging.

*Some of the mails that seem not to send appear when I relog. It seems they can't be deleted.

I feel like these issues stem from how the expire time is currently implemented in my proposed changes, or maybe my changes fucked something up somewhere. **It's worth to note that I added a line in Mail.java `public long expireTime;` at line 70 as the code would not compile otherwise.**

**tl;dr the command works properly, the expireTime is written to the database properly. I believe the issue is with getting data from the database.** 

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -102,6 +102,8 @@ public final class Grasscutter {
 	
 		// Initialize server.
 		Grasscutter.getLogger().info(translate("messages.status.starting"));
+		Grasscutter.getLogger().info(translate("messages.status.game_version", GameConstants.VERSION));
+		Grasscutter.getLogger().info(translate("messages.status.version", BuildConfig.VERSION, BuildConfig.GIT_HASH));
 	
 		// Load all resources.
 		Grasscutter.updateDayOfWeek();
--- a/src/main/java/emu/grasscutter/auth/DefaultAuthenticators.java
+++ b/src/main/java/emu/grasscutter/auth/DefaultAuthenticators.java
@@ -36,7 +36,7 @@ public final class DefaultAuthenticators {
                 // Check if account exists.
                 if(account == null && ACCOUNT.autoCreate) {
                     // This account has been created AUTOMATICALLY. There will be no permissions added.
-                    account = DatabaseHelper.createAccountWithId(requestData.account, 0);
+                    account = DatabaseHelper.createAccountWithUid(requestData.account, 0);
 
                     // Check if the account was created successfully.
                     if(account == null) {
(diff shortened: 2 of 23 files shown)


AZthemute:
In mail.java, I made a few changes:
`public static long var = 86400;`
`public Mail() {
        this(new MailContent(), new ArrayList<MailItem>(), (int) Instant.now().getEpochSecond() + var);
    }`
Instead of a hardcoded value, it now uses a hardcoded variable to decide the expiry time.

image
image
Everything works as expected.

Now, let's do the same thing but with the expireTime variable:
`public static long expireTime = 86400;` (it has to be static or else the program does not compile)
`public Mail() {
        this(new MailContent(), new ArrayList<MailItem>(), (int) Instant.now().getEpochSecond() + expireTime);
    }`

image
image
There's nothing in the database and the expiry time is... quite far off.

It's the same principle of using a variable to represent the amount of time until expiry, however for some reason this name is just... cursed. I may do the dirty solution of, well, what I did just before this. But obviously with better variable names. I just need to figure out how to pass the variable to this, but I think this is the best option because I'm too lazy to look for another solution, and it works fine anyway.

AZthemute:
Don't know what the fuck I did, but getting a clean copy of my commits and doing what I did now um...
image
No clue why. It's that time again when we have to reset...

...unless someone wants to fix this, but I doubt it. The command is there but actually getting this to do anything is a little more complicated.

PR #1194, opened 2022-06-04
Fix unable to save game data occasionally

zhaodice:
## Fix bugs:

**1. Max players limit**

If there are 5 online players, max count of player is 5,
a new client ( who has one of their account) want to login by kicking one of them ,
I think it should be allowed

**2. Check whether user exists, if exists, kick it and save data( save immediately! )**

Checking must situated here (before `getPlayerByUID`)! 
because to save firstly ,to load secondly !!!
or the `getPlayerByUID` will load old data

**3.Remove offline player from server**

Don't delete by uid,because there are multiple same uid players when duplicated login!
so I decide to delete by object rather than uid

**4.`exitDungeon` added a sanity check in case of null error**

**5.Fix data lost**
Fix when player exit game or kicked by others login, client quickly login again will lost recently record.

I am sorry but too many problem with this project, I can't edit simply, if you have any question ,feel free to ask me!

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/server/game/GameServerInitializer.java
+++ b/src/main/java/emu/grasscutter/server/game/GameServerInitializer.java
@@ -13,6 +13,10 @@ public class GameServerInitializer extends KcpServerInitializer {
 	
     @Override
     protected void initChannel(UkcpChannel ch) throws Exception {
-        new GameSession(server,ch.pipeline());
+        ChannelPipeline pipeline=null;
+        if(ch!=null){
+            pipeline = ch.pipeline();
+        }
+        new GameSession(server,pipeline);
     }
 }
--- a/src/main/java/emu/grasscutter/server/game/GameServerInitializer.java
+++ b/src/main/java/emu/grasscutter/server/game/GameServerInitializer.java
@@ -13,8 +13,6 @@ public class GameServerInitializer extends KcpServerInitializer {
 	
     @Override
     protected void initChannel(UkcpChannel ch) throws Exception {
-        ChannelPipeline pipeline = ch.pipeline();
-        GameSession session = new GameSession(server);
-        pipeline.addLast(session);
+        new GameSession(server,ch.pipeline());
     }
 }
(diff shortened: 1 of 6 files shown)


zhaodice:
I am sorry but too many problem with this project, I can't edit simply, if you have any question ,fell free to ask me

zhaodice:
All done, I have tested it ,it would be very stable

zhaodice:
Note : commit `7b07ad6` was delected by `e2db401`

src/main/java/emu/grasscutter/server/game/GameServerInitializer.java
```diff
@@ -13,8 +13,6 @@ public GameServerInitializer(GameServer server) {
 	
     @Override
     protected void initChannel(UkcpChannel ch) throws Exception {
-        ChannelPipeline pipeline = ch.pipeline();
-        GameSession session = new GameSession(server);
-        pipeline.addLast(session);
+        new GameSession(server,ch.pipeline());
```
ghost: Please keep the existing code as is, in case a session needs to be created without a pipeline.

src/main/java/emu/grasscutter/server/packet/recv/HandlerGetPlayerTokenReq.java
```diff
@@ -20,28 +21,45 @@ public class HandlerGetPlayerTokenReq extends PacketHandler {
...
-		if (ACCOUNT.maxPlayer > -1 && Grasscutter.getGameServer().getPlayers().size() >= ACCOUNT.maxPlayer) {
-			session.close();
-			return;
-		}
-		
+
 		GetPlayerTokenReq req = GetPlayerTokenReq.parseFrom(payload);
 		
 		// Authenticate
-		Account account = DatabaseHelper.getAccountById(req.getAccountUid());
```
ghost, line 32: Please keep the previous code here OR do an account id check just in case.

ghost (changes requested):

zhaodice:
All done.

AZthemute:
> If there are 5 online players, max count of player is 5,
> a new client want to login by kicking one of them ,
> I think it should be allowed

As an option we can easily toggle?

zhaodice:
> > If there are 5 online players, max count of player is 5,
> > a new client want to login by kicking one of them ,
> > I think it should be allowed
> 
> As an option we can easily toggle?

do you mean you want a switch that when no available position, could decline new client's login request although the client has one of their account?

zhaodice:
> As an option we can easily toggle?

If there are 5 online players, max count of player is 5,
a new client **( who has one of their account)** want to login by kicking one of them ,
I think it should be allowed

ghost (approved):

PR #1236, opened 2022-06-10
Fix EquipInventoryTab.getItemById always returning null

ffauzan:
## Description

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


ghost (changes requested):
Equippable items are unique and a player can have duplicate items with the same item id but with different stats, therefore, they should be NOT be in a hashmap with their item id as their key.

ffauzan:
> Equippable items are unique and a player can have duplicate items with the same item id but with different stats, therefore, they should be NOT be in a hashmap with their item id as their key.

I see, so getting an equipable item by its item id would return a set of items & that can be achieved by just stream filter the inventory. Thanks for the explanation, I will close this pr.

PR #1245, opened 2022-06-12
Command: Add permissionTargeted or targetRequirement

Tesutarin:
## Description
Command: Add permissionTargeted or targetRequirement
## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/KickCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/KickCommand.java
@@ -8,7 +8,7 @@ import java.util.List;
 
 import static emu.grasscutter.utils.Language.translate;
 
-@Command(label = "kick", usage = "kick", permission = "server.kick", description = "commands.kick.description")
+@Command(label = "kick", usage = "kick", permission = "server.kick", permissionTargeted = "server.kick.others", description = "commands.kick.description")
 public final class KickCommand implements CommandHandler {
 
     @Override
--- a/src/main/java/emu/grasscutter/command/commands/PermissionCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/PermissionCommand.java
@@ -10,7 +10,7 @@ import java.util.List;
 
 import static emu.grasscutter.utils.Language.translate;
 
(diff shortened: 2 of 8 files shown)


ghost (approved):

Birdulon:
PermissionCommand and PositionCommand require targets and will fail without them. You can verify this by testing them from the server console without setting a target, or simply reading their functions and seeing if they use `targetPlayer` for anything.

src/main/java/emu/grasscutter/command/commands/PermissionCommand.java
```diff
@@ -10,7 +10,7 @@
 
 import static emu.grasscutter.utils.Language.translate;
 
-@Command(label = "permission", usage = "permission <add|remove> <permission>", permission = "permission", description = "commands.permission.description")
+@Command(label = "permission", usage = "permission <add|remove> <permission>", permission = "server.permission", description = "commands.permission.description", targetRequirement = Command.TargetRequirement.NONE)
```
Birdulon, line 13: This is a very dangerous change as server operators tend to hand out `server.*` permissions. Negative permissions make this an easy config fix but this would need to be **very** visible on the changelog.

src/main/java/emu/grasscutter/command/commands/KickCommand.java
```diff
@@ -8,7 +8,7 @@
 
 import static emu.grasscutter.utils.Language.translate;
 
-@Command(label = "kick", usage = "kick", permission = "server.kick", description = "commands.kick.description")
```
Birdulon, line 11: Kick is one of those commands where it didn't really make sense to have an `others` subpermission as the whole point of the command is to use on other players. `/restart` actually works identically to using `/kick` on oneself. If `kick` is to gain separate permissions for self vs other usage, `restart` may as well be deprecated.

Tesutarin:
> PermissionCommand and PositionCommand require targets and will fail without them. You can verify this by testing them from the server console without setting a target, or simply reading their functions and seeing if they use `targetPlayer` for anything.

got it, thanks for ur explanation

PR #1251, opened 2022-06-13
Separating dispatch server and game server

exzork:
## Description

## Issues fixed by this PR

Separating dispatch server and game server. 
Now dispatch server can use only one database ( one account for multiple game server ) and each game server have their own database. 
Dispatch and game server will be communicating using websocket.

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


zhaodice:
I think dispatch server needn't communicate with game server.
dispatch just need encrypt account information(password and username,uid) with RSA into token,and game server decrypt and verify

exzork:
> I think dispatch server needn't communicate with game server. dispatch just need encrypt account information(password and username,uid) with RSA into token,and game server decrypt and verify

Some of function/method is using account object as their parameter/return so without communicating i don't think it's possible to separate this. Well, you can but it needs to modify that function/method

mcxiaoke:
why websocket, why not just tcp (grpc/thrift)? or http?

exzork:
> why websocket, why not just tcp (grpc/thrift)? or http?

Well, it's just my idea of using websocket. Because we don't need to open new port ( http api is the same tho ).

ghost:
No need to use websockets or anything to bridge the game/dispatch servers together, ideally you should seperate the datastores into an accounts and game datastore which can connect to different mongodb instances if set in the config. The dispatch server would use only the account datastore, and the game server would use both datastores.

exzork:
> No need to use websockets or anything to bridge the game/dispatch servers together, ideally you should seperate the datastores into an accounts and game datastore which can connect to different mongodb instances if set in the config. The dispatch server would use only the account datastore, and the game server would use both datastores.

It's already like that as of grasscutter now right? 
On this pr ( as for i write this comment ) we can set "dispatch key" and the dispatch server will open websocket for the game client to connect to and thus updating available game server ( on game client server selection ). Uid and player counter are handled in dispatch server so no uid conflict will happen.

4Benj:
> No need to use websockets or anything to bridge the game/dispatch servers together, ideally you should seperate the datastores into an accounts and game datastore which can connect to different mongodb instances if set in the config. The dispatch server would use only the account datastore, and the game server would use both datastores.

In my opinion having two servers interact with the same database is a bad idea and I only did it that way just to get it working when I separated the servers and the web socket approach was something that I had planned on submitting a PR for later but Exzork obviously done it I could.

exzork:
Testing server is on my server, for address you can check it on my discord profile ( ExZork#6937 ). Currently has 2 regions. SG is hybrid with dispatch and USA is game only.

exzork:
I'm planning to rewrite and restructure this shit first before it's ready for review

exzork:
I'll rewrite this using socket io instead of plain websocket

PR #1254, opened 2022-06-13
Update Epitomized Path

CamChuaVN:
## Description
Update epitomized path to gacha system

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


src/main/java/emu/grasscutter/game/gacha/GachaManager.java
```diff
@@ -183,11 +184,28 @@ private synchronized int drawRoulette(int[] weights, int cutoff) {
 
 	private synchronized int doRarePull(int[] featured, int[] fallback1, int[] fallback2, int rarity, GachaBanner banner, PlayerGachaBannerInfo gachaInfo) {
 		int itemId = 0;
-		boolean pullFeatured = (gachaInfo.getFailedFeaturedItemPulls(rarity) >= 1)  // Lost previous coinflip
+		boolean pullFeatured = (gachaInfo.getFailedFeaturedItemPulls(rarity) >= (banner.hasEpitomized() ? banner.getWishMaxProgress() : 1))  // Lost previous coinflip
```
Birdulon: 1) This appears to be missing a `rarity` check
2) I think this breaks the guarantee of getting a featured item after losing a coinflip/50:50 - need another counter instead of reusing `FailedFeaturedItemPulls`. Maybe it could be called `FailedChosenItemPulls`?

It's probably cleaner to move the check for `WishMaxProgress` to the start, and check if the item matches at the end, something like this:
```java
...{
int itemId = 0;
boolean epitomized = (banner.hasEpitomized() && (rarity == 5));
if (epitomized && (gachaInfo.getFailedChosenItemPulls() >= banner.getWishMaxProgress())) {
    itemId = gachaInfo.getWishItemId(); // Pick the chosen item
} else {
    boolean pullFeatured = ...
    if (pullFeatured && (featured.length > 0)) {
        // Pick a featured item
    } else {
        // Pick a fallback item
    }
}
if (epitomized && (itemId != gachaInfo.getWishItemId())
    gachaInfo.setFailedChosenItemPulls(0);
else
    gachaInfo.addFailedChosenItemPulls(1);

return itemId;
}
```

CamChuaVN: okay thanks, i will update soon

src/main/java/emu/grasscutter/server/packet/recv/HandlerGetGachaInfoReq.java
```diff
@@ -15,6 +18,13 @@ public void handle(GameSession session, byte[] header, byte[] payload) throws Ex
 					// TODO: use other Nonce/key insteadof session key to ensure the overall security for the player
 					session.getPlayer().getAccount().getSessionKey())
 					);
+
+		for(GachaBanner banner : session.getServer().getGachaManager().getGachaBanners().values()) {
+			if(!banner.hasEpitomized()) continue;
+
+			PlayerGachaBannerInfo gachaInfo = session.getPlayer().getGachaInfo().getBannerInfo(banner);
+			session.send(new PacketGachaWishRsp(banner.getGachaType(), banner.getScheduleId(), gachaInfo.getWishItemId(), gachaInfo.getFailedFeaturedItemPulls(5), banner.getWishMaxProgress()));
```
GanyusLeftHorn: This seems strangely placed here. We don't usually send an *Rsp without a preceeding matching *Req. This should probably only be sent in response to GachaWishReq, including the information about failed pulls (which is currently always set to zero there).

CamChuaVN: If dont send Rsp on this, when click to weapon banner, it dont show current wished progress

GanyusLeftHorn: You are explicitly setting the wish progress to zero when replying to GachaWishReq. That is probably why you see no progress if you don't add it here. Try setting the correct progress in HandleGachaWishReq.

CamChuaVN: GachaWishReq only called when click Chart Course or Cancel Course, i set wish progress to zero because when Cancel Course will reset current progress, when click to banner it only call GetGachaInfoRsp and will not show current wish progress, i add GachaWishRsp in GetGachaInfoRsp will make client see current wish progress, watch my video below
[HandlerGetGachaInfoReq with GachaWishRsp.zip](https://github.com/Grasscutters/Grasscutter/files/8898556/HandlerGetGachaInfoReq.with.GachaWishRsp.zip)
[HandlerGetGachaInfoReq without GachaWishRsp.zip](https://github.com/Grasscutters/Grasscutter/files/8898557/HandlerGetGachaInfoReq.without.GachaWishRsp.zip)

GanyusLeftHorn: Ah, yes, because `wish_item_id`, `wish_progress` and `wish_max_progress` are not set in `GetGachaInfoRsp`. Have a look at `GachaBanner::toProto(String)`.

CamChuaVN: i have updated

Birdulon (approved):

ghost (approved):

PR #1290, opened 2022-06-17
Account datastore should be managed by config.json, not a hardcode

f0re1gnKey:
## Description
Account datastore should be managed by config.json, not a hardcode.
By default, the config.json use the same datastore.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


GanyusLeftHorn:
`dispatchDatastore` is not initialized outside of runmode `GAME_ONLY`. I haven't tested it, but from what I can tell, this change should make the server crash for example when trying to create an account while in a different runmode.

f0re1gnKey:
> `dispatchDatastore` is not initialized outside of runmode `GAME_ONLY`. I haven't tested it, but from what I can tell, this change should make the server crash for example when trying to create an account while in a different runmode.

Sorry for making a mistake.

f0re1gnKey:
> `dispatchDatastore` is not initialized outside of runmode `GAME_ONLY`. I haven't tested it, but from what I can tell, this change should make the server crash for example when trying to create an account while in a different runmode.

Sorry for making a mistake.

4Benj:
In it's current state, a server in ``GAME_ONLY`` mode needs access to two databases, one managed by the dispatch server and one managed by itself (which I very much dislike as seen by my comment but it's usable for now), which #1251 aims to fix.
Until/If #1251 gets merged, the current solution is the only viable option imo because a server running in ``GAME_ONLY`` mode cannot use the same datastore as the dispatch server unless the dispatch server is ``DISPATCH_ONLY`` server and the only other server connected is the previously mentioned ``GAME_ONLY`` server. 

Honestly, I'm not entirely sure what you're trying to do here, so I'm just gonna close this. Feel free to explain in more detail your goal and this can be reopened if it seems reasonable.

PR #1341, opened 2022-06-21
Fix talent material cannot be added when wish

Chika99:
## Description

Fix when more than two duplicate characters are drawn in wish, the talent material of the character will not increase.
## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


Birdulon:
If this is about the `stackLimit`s bug that was addressed in #1333, it's not applicable in current development branch since we reset to before the commit which introduced it. Try reproducing on the current development branch, there should be no need to implement a workaround for item `stackLimit`s.

Chika99:
Oh I just found this, I re-tested and it was indeed a `stackLimit` bug.

PR #1345, opened 2022-06-22
Update weather command

Tesutarin:
## Description
A simpler way to change the weather and climate
`weatherId` is the opposite of `climateId`. `climateId` stands for current weather
now you can just type `/w 0` to change the weather to `none`, etc
## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


Birdulon:
I still don't understand why you want to set the climate without setting the weather, when the weather should control the climate.

PR #1444, opened 2022-07-05
Added more server debug options

akatatsu27:
## Description
In config.json on "debugLevel" you can now also input "WHITELIST" or "BLACKLIST".
The console will display or blacklist any packets (including those excluded by ALL: PingReq, PingRsp, WorldPlayerRTTNotify, UnionCmdNotify and QueryPathReq) whose opcodes are in the new fields "DebugWhitelist", and "DebugBlacklist" respectively. Did not touch the version number in ConfigContainer.java
## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x ] Enhancement
- [ ] Documentation

## Checklist:

- [x ] My code follows the style guidelines of this project
- [ x] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [x ] I am responsible for any copyright issues with my code if it occurs in the future.


src/main/java/emu/grasscutter/server/game/GameSession.java
```diff
@@ -150,6 +150,32 @@ public void send(BasePacket packet) {
 				System.out.println(Utils.bytesToHex(packet.getData()));
 			}
     	}
+    	
+    	if (SERVER.debugLevel == ServerDebugMode.WHITELIST) {
+    		for (int i=0; i < SERVER.serverDebugWhitelist.length; i++) {
+    			if (packet.getOpcode() == SERVER.serverDebugWhitelist[i]) {
```
GanyusLeftHorn: This is way more complicated than it has to be. For example, if you store your white- and blacklist as lists instead of arrays, this simply becomes
```java
if (SERVER.debugLevel == ServerDebugMode.WHITELIST && SERVER.serverDebugWhitelist.contains(opcode)) {
```

src/main/java/emu/grasscutter/server/game/GameSession.java
```diff
@@ -220,6 +246,31 @@ public void handleReceive(byte[] bytes) {
...
+						if (opcode == SERVER.serverDebugBlacklist[i]) {
+		    				beBlacklisted = true;
+		    				break;
+						}		
+					}
+					if (!beBlacklisted) {
+						Grasscutter.getLogger().info("RECV: " + PacketOpcodesUtil.getOpcodeName(opcode) + " (" + opcode + ")");
+						System.out.println(Utils.bytesToHex(payload));
+					}
+				}
```
GanyusLeftHorn: The logging code here is pretty much identical to the one in `send`. We should probably factor that out into a separate method and just call that in both `send` and  `handleReceive`.

src/main/java/emu/grasscutter/server/game/GameSession.java
```diff
@@ -150,6 +150,32 @@ public void send(BasePacket packet) {
...
+    	if (SERVER.debugLevel == ServerDebugMode.BLACKLIST) {
+    		boolean beBlacklisted = false;
+     		for (int i=0; i < SERVER.serverDebugBlacklist.length; i++) {
+    			if (packet.getOpcode() == SERVER.serverDebugBlacklist[i]) {
+    				beBlacklisted = true;
+    				break;
+    			}
+    			
+     		}
+     		if (!beBlacklisted) {
```
GanyusLeftHorn: Likewise, this would become something like

```java
if (SERVER.debugLevel == ServerDebugMode.BLACKLIST && !SERVER.serverDebugBlacklist.contains(opcode)) {
```

GanyusLeftHorn (changes requested):

GanyusLeftHorn:
I don't know how necessary adding both a whitelist and blacklist functionality is. I have definitely wanted the ability to log _everything_, including things like `UnionCmd`, before, so this added functionality is appreciated (at least for me personally). But whether we actually need both, I don't know.

src/main/java/emu/grasscutter/server/game/GameSession.java
ghost: They should be stored as hashsets, not arrays.

src/main/java/emu/grasscutter/utils/ConfigContainer.java
```diff
@@ -93,6 +93,8 @@ public static class Structure {
 
     public static class Server {
         public ServerDebugMode debugLevel = ServerDebugMode.NONE;
+        public int[] serverDebugWhitelist = {};
```
ghost: No need to append "server" in front of the config variable name.

ghost (changes requested):

ghost (approved):

PR #1479, opened 2022-07-09
Fully Implement OpenState framework

akatatsu27:
## Description

- Fully implements the OpenState framework.
- Sets all OpenStates
- Correctly sends/handles packets: SetOpenStateChangeNotify, OpenStateUpdateNotify, SetOpenStateReq, SetOpenStateRsp
## Type of changes
- [ ] Bug fix
- [x ] New feature 
- [x ] Enhancement
- [ ] Documentation

## Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [x ] I am responsible for any copyright issues with my code if it occurs in the future.


src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+    public void setPlayer(Player player) {
+        this.player = player;
+    }
+
+    public Integer getOpenState(OpenState openState) {
+        if(!openStateMap.containsKey(openState)){
+            //default to 0
+            openStateMap.putIfAbsent(openState.getValue(),0);
+        }
+        return this.openStateMap.get(openState);
```
GanyusLeftHorn: 1) This will never work - `openStateMap` is a `Map<Integer, Integer>`, but this code uses `OpenState` as key to both check for existance as well as fetching the value, but crucially not for insertion. TL/DR: This method always returns `null`. Which you won't notice, since it is never called anyway.
1) The `if` here is unnecessary - `putIfAbsent` only writes the value if it is not already in the map. 
1) That being said, I don't think it is even necessary to insert the state into the map if it is missing. Simply use 
```
return this.openStateMap.getOrDefault(openState.getValue(), 0);
```

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    public void setOpenState(OpenState openState, Integer value) {
+        if(!openStateMap.containsKey(openState.getValue())){
+            openStateMap.putIfAbsent(openState.getValue(),value);
+        } else {
+            this.openStateMap.put(openState.getValue(), value);
+        }
```
GanyusLeftHorn: This is equivalent to 
```
this.openStateMap.put(openState.getValue(), value);
```

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+        player.getSession().send(new PacketOpenStateChangeNotify(openState.getValue(),value));
```
GanyusLeftHorn: Don't send a packet on every call to this method. Only send it if the value actually _changed_. Sending it every time leads to the player being barraged with unlock notifications on every login (starting from the second, but more on that later).

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public void setOpenState(int openState, Integer value) {
+        if(!openStateMap.containsKey(openState)) {
+            openStateMap.putIfAbsent(openState,value);
+        } else {
+            this.openStateMap.put(openState, value);
+        }
+        player.getSession().send(new PacketOpenStateChangeNotify(openState,value));
+    }
```
GanyusLeftHorn: This overload seems unnecessary. `setOpenState(OpenState, Integer)` should be enough.

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+        for(Map.Entry<OpenState, Integer> entry : openStatesChanged.entrySet()) {
+            setOpenState(entry.getKey(), entry.getValue());
+        }
+    }
+
+    public void setOpenStatesByKeyValue(Map<Integer,Integer> openStatesChanged) {
+        for(Map.Entry<Integer, Integer> entry : openStatesChanged.entrySet()) {
+            setOpenState(entry.getKey(), entry.getValue());
+        }
+    }
```
GanyusLeftHorn: Those two methods are essentially equivalent (and have misleading names - both of them set both key and value). One of them should be enough.

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public void onNewPlayerCreate() {
+       Set<OpenState> newPlayerOpenStates = Set.of(OPEN_STATE_DERIVATIVE_MALL,OPEN_STATE_PHOTOGRAPH,OPEN_STATE_BATTLE_PASS,OPEN_STATE_SHOP_TYPE_GENESISCRYSTAL,OPEN_STATE_SHOP_TYPE_RECOMMANDED,
+           OPEN_STATE_SHOP_TYPE_GIFTPACKAGE,OPEN_STATE_GUIDE_RELIC_PROM,OPEN_STATE_GUIDE_TALENT,OPEN_STATE_SHOP_TYPE_BLACKSMITH,OPEN_STATE_SHOP_TYPE_PAIMON,OPEN_STATE_WEAPON_AWAKEN,
+           OPEN_STATE_WEAPON_PROMOTE,OPEN_STATE_AVATAR_PROMOTE,OPEN_STATE_AVATAR_TALENT,OPEN_STATE_WEAPON_UPGRADE,OPEN_STATE_RESIN,OPEN_STATE_RELIQUARY_UPGRADE,
+           OPEN_STATE_SHOP_TYPE_VIRTUAL_SHOP,OPEN_STATE_RELIQUARY_PROMOTE);
+
```
GanyusLeftHorn: This should probably be a static member of the class.

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+        Map<OpenState,Integer> openStatesChanged = new HashMap<>();
+        Iterator<OpenState> iterator = newPlayerOpenStates.iterator();
+        while (iterator.hasNext()) {
+            openStatesChanged.put(iterator.next(), 1);
+        }
+        setOpenStatesByKey(openStatesChanged);
```
GanyusLeftHorn: Shorter:
```
newPlayerOpenStates.forEach(os -> this.setOpenState(os, 1));
```

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    public void onPlayerLogin() {
+        //little hack to give all openStates on second login
+        if(openStateMap.containsKey(16)) {
+            setAllOpenStates();
+        }
+        player.getSession().send(new PacketOpenStateUpdateNotify(player));
```
GanyusLeftHorn: This is where I get really confused. Why set all OpenStates to 1 starting from the second login? Why even track them in the database at all then? Our current behavior is to set all OpenStates to 1 for _every_ login. Doing it for every login starting from the second is a marginal difference (which only leads to inconvenience for the player, with various systems like crafting or wishing only accessible starting from the second login).

This defeats the whole purpose of actually tracking OpenStates (which is a good thing that we definitely want to do in the long run, don't get me wrong). The correct thing to do would be to 
1) On _every_ login, return exactly the OpenStates stored in the database.
1) Implement all the ways in which OpenStates can get set (crucially including quest triggers).

Unfortunately, given that quests are not fully functional yet, this is probably going to be a problem. I would therefore suggest turning this PR into more of an infrastructure thing: Add OpenState tracking to the database, expose an interface for getting and setting them, _but_ be very generous with which OpenStates we set right off the bat (read: set all of them, as we currently do, or at least almost all of them). We can then gradually reduce the list of automatically set OpenStates when we add handling for the methods in which they are actually set.

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public void setAllOpenStates() {
+        Map<Integer,Integer> openStatesChanged = new HashMap<>();
+        List<Integer> allOpenStates = Stream.of(OpenState.values())
+            .map(OpenState::getValue).collect(Collectors.toList());
+        for(int i=0; i<allOpenStates.size(); i++) {
+            openStatesChanged.put(allOpenStates.get(i), 1);
+        }
+        setOpenStatesByKeyValue(openStatesChanged);
```
GanyusLeftHorn:
```
Stream.of(OpenState.values()).forEach(os -> this.setOpenState(os, 1));
```

GanyusLeftHorn (changes requested):

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
akatatsu27: I was comparing with official server conversation with a new Player, and during the first quest, the client sends some SetOpenStateReq packets. I think its crucial we try to mimic the official server conversation to get stuff working. I've changed it to set all OpenStates, but I'll leave the original onNewPlayerCreate commented out, for anyone who wants to experiment with new Player initial conditions

PR #1483, opened 2022-07-10
Open state framework

akatatsu27:
## Description
Fully implements the OpenState Framework
-Correctly Handles/receives packets: SetOpenStateReq, SetOpenStateRsp, OpenStateChangeNotify, OpenStateUpdateNotify
-OpenStates of player saved to/loaded from the database. (if you want to add OpenStates to an existing Player, either edit the player object in the database, or edit the code to make PlayerOpenStateManager::onPlayerLogin set devOpenStates, and make sure to revert your changes afterwards)
-Sets all OpenStates in devOpenStates to 1 on new Player creation 
-Debug message if server (or client) tries to set an OpenState to the same value 
-Made a Set of OpenStates that should be set on new Player creation, based on official server behaviour (can be found in PlayerOpenStateManager)
## Type of changes
- [ ] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


akatatsu27:
Sorry for making a new PR. I was having trouble with my IDE, and the Files Changed tab was a mess. I made the changes GanyusLeftHorn suggested in the original PR

src/main/java/emu/grasscutter/game/player/PlayerOpenStateManager.java
```diff
@@ -0,0 +1,74 @@
...
+@Entity
+public class PlayerOpenStateManager {
+
+    @Transient private Player player;
+    @Getter private Map<Integer,Integer> openStateMap;
+    //DO NOT MODIFY. Based on conversation of official server and client, game version 2.7
+    private static Set<OpenState> newPlayerOpenStates = Set.of(OPEN_STATE_DERIVATIVE_MALL,OPEN_STATE_PHOTOGRAPH,OPEN_STATE_BATTLE_PASS,OPEN_STATE_SHOP_TYPE_GENESISCRYSTAL,OPEN_STATE_SHOP_TYPE_RECOMMANDED,
+        OPEN_STATE_SHOP_TYPE_GIFTPACKAGE,OPEN_STATE_GUIDE_RELIC_PROM,OPEN_STATE_GUIDE_TALENT,OPEN_STATE_SHOP_TYPE_BLACKSMITH,OPEN_STATE_SHOP_TYPE_PAIMON,OPEN_STATE_WEAPON_AWAKEN,
+        OPEN_STATE_WEAPON_PROMOTE,OPEN_STATE_AVATAR_PROMOTE,OPEN_STATE_AVATAR_TALENT,OPEN_STATE_WEAPON_UPGRADE,OPEN_STATE_RESIN,OPEN_STATE_RELIQUARY_UPGRADE,
+        OPEN_STATE_SHOP_TYPE_VIRTUAL_SHOP,OPEN_STATE_RELIQUARY_PROMOTE);
```
GanyusLeftHorn, line 28: I suggest turning this into a set of _all_ the OpenStates we want to set on player creation. In other words, replace to call to `setAllOpenStates` with the currently commented out code, and turn this set into a full set of all the states. This would make it easier to gradually remove OpenStates from the set once we implement the triggers that set them. 

And since you already implemented `SetOpenStateReq`, you could probably remove all the OpenStates related to UI hints from the set. Those get set via `SetOpenStateReq` once the client has displayed them, if I am not mistaken.

akatatsu27: these specific openStates are the ones the official server set when I made a new player, so I think we should keep these as is for anyone who wants to investigate/experiment with new player initial conditions. Alternatively, we should make a second list of the openStates we currently want to set, without touching newPlayerOpenStates.

GanyusLeftHorn: You can put them in a comment, if you want to document what the states that get sent on official are. But making an unused member for them is pointless, imo.

GanyusLeftHorn:
Two things remaining:
1) Please remove `OPEN_STATE_NONE` from the set. That was not set in our current implementation.
1) I don't think we need to print a warning every time an already set OpenState is attempted to be set again. That will probably spam the console?

As I mentioned before, given that you already implemented `SetOpenStateReq`, we could probably remove some guide-related OpenStates from the set and let them be set naturally, but I guess that is also something that we can do in a later PR.

Aside from the points above, this looks good to me now.

akatatsu27:
The debug message won' spam the console, because we only send OpenStateUpdateNotify on login. Besides, it should aid anyone implementing openState triggers in the future.  Lastly, I tried removing any openState that had the word GUIDE in it, but the client sent me a bunch of SetOpenStateReq, I had some UI guides that wouldn't go away, and I even saw my debug message appear. I think each openState should be addressed by a relevant PR

GanyusLeftHorn:
Alright, looks good to me!

ghost (approved):

PR #1490, opened 2022-07-11
[Bugfix] Multi-server mode

ITer2022:
## Description
The old getAccountDatastore() doesn't work properly when setting to multi-server mode. Always two datastore, one is dispatchDatastore, which is shared by all servers, and the other one is gameDatastore, which is seperated.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/database/DatabaseManager.java
+++ b/src/main/java/emu/grasscutter/database/DatabaseManager.java
@@ -49,11 +49,11 @@ public final class DatabaseManager {
 	// Yes. I very dislike this method. However, this will be good for now.
 	// TODO: Add dispatch routes for player account management
 	public static Datastore getAccountDatastore() {
-		if(SERVER.runMode == ServerRunMode.GAME_ONLY) {
+		//if(SERVER.runMode == ServerRunMode.GAME_ONLY) {
 			return dispatchDatastore;
-		} else {
-			return gameDatastore;
-		}
+		//} else {
+		//	return gameDatastore;
+		//}
 	}
 	
 	public static void initialize() {
@@ -85,7 +85,7 @@ public final class DatabaseManager {
 			}
 		}
 
-		if(SERVER.runMode == ServerRunMode.GAME_ONLY) {
+		//if(SERVER.runMode == ServerRunMode.GAME_ONLY) {
 			MongoClient dispatchMongoClient = MongoClients.create(DATABASE.server.connectionUri);
 			dispatchDatastore = Morphia.createDatastore(dispatchMongoClient, DATABASE.server.collection);
 
@@ -105,7 +105,7 @@ public final class DatabaseManager {
 					dispatchDatastore.ensureIndexes();
 				}
 			}
-		}
+		//}
 	}
 
 	public static synchronized int getNextId(Class<?> c) {


ITer2022:
account info --> dispatchDatastore (shared by all servers)
game info --> gameDataStore (gs1, gs2, ... seperated)
TESTING PASSED. How to set to multi-server mode? [Click here.](https://www.cnblogs.com/ITer2022/p/16462280.html)

4Benj:
imo There should never be two datastore and multiple servers should not share the same database. However, due to my own laziness and the state of which the server was at when I implemented this, making a separate datastore was the easiest solution and has always meant to be a temporary solution since having two servers interact with the same database can cause some problems (although rare). Currently the plan is to completely separate the servers put into game and dispatch mode and just have them communicate with each other via web sockets. 
I do not think this is the right solution to solving this issue tbh and while it would be a fix for the current problem, it's a temporary one.

May I ask what exactly are the bug fixes that this intends to fix? Is there any difficulty or errors setting up separated servers without this fix?

ITer2022:
Well, in official game, we can enter different regions like usa, eur... with the same account. And the game data of different regions are independent of each other.
Without this fix, we can not reach that goal.

4Benj:
Yes we can...? and that's how it works at the moment
The dispatch datastore is a temporary fix which should only connect to the database that stores account information so that the game server can verify that the users account exist, but everything game related is stored on the game's datastore.
The dispatch datastore **should only be used in game only mode** so the game server can verify the account so it doesn't create any unnecessary overhead in other server run modes where it is not necessary.

4Benj:
I've seen multiple servers who have their game data split up already between regions.

ITer2022:
Suppose there are two servers.
1. We set one to HYBRID or DISPATCH_ONLY mode and the other to GAME_ONLY mode. The function getAccountDatastore() returns gameDatastore on HYBRID-mode server, but dispatchDatasotre on GAME_ONLY-mode server, witch will cause white door error, cuz GAME_ONLY-mode server can not fetch account info.
2. Now, we set both to HYBRID mode. The function getAccountDatastore() returns the same gameDatastore. We enter the same region actually even though we could choose regions on client.

ITer2022:
Suppose there are two servers.
1. We set one to HYBRID or DISPATCH_ONLY mode and the other to GAME_ONLY mode. The function getAccountDatastore() returns gameDatastore on HYBRID-mode server, but dispatchDatasotre on GAME_ONLY-mode server, witch will cause white door error, cuz GAME_ONLY-mode server can not fetch account info.
2. Now, we set both to HYBRID mode. The function getAccountDatastore() returns the same gameDatastore. We enter the same region actually even though we could choose regions on client.

ITer2022:
> I've seen multiple servers who have their game data split up already between regions.

Yes, I've ever seen too. When i tried to login, i got the same UID in different regions.

4Benj:
> Yes, I've ever seen too. When i tried to login, i got the same UID in different regions.

So? Why does the account uid matter? This change won’t fix that and to fix that will require changing how grasscutter handles uids which doesn’t make sense to do. Having the same uid makes the account easily identifiable for server administrators and the region doesn’t matter.
It doesn’t make sense that in official the user ids are different anyway.

ITer2022:
Not the account uid, but the player uid. The amount of players is not the same in diffrent regions, when a newly registered account login, won't it get diffrent palyer uid in the lower right coner between regions?

4Benj:
> Not the account uid, but the player uid. The amount of player is not the same in diffrent regions, when a newly registered account login, won't it get diffrent palyer uid in the lower right coner between regions?

Sorry, my mistake, I did mean player uid but that doesn't change the fact that having different player uid for each region doesn't make sense and that it doesn't even matter anyway since the two game servers should not share the same database.

4Benj:
> Suppose there are two servers.
> 
> 1. We set one to HYBRID or DISPATCH_ONLY mode and the other to GAME_ONLY mode. The function getAccountDatastore() returns gameDatastore on HYBRID-mode server, but dispatchDatasotre on GAME_ONLY-mode server, witch will cause white door error, cuz GAME_ONLY-mode server can not fetch account info.
> 2. Now, we set both to HYBRID mode. The function getAccountDatastore() returns the same gameDatastore. We enter the same region actually even though we could choose regions on client.

That is an issue, I will look into it but you have solved the solution but in the wrong way which will cause issues in the future.

ITer2022:
Sorry, too. I didn't make it clear. I use the same UID to indicate that the players we created are the same between regions. Suppose we've already created a player in region1, when we tried to login to region2, we got the same player we created in region1, not a newly created.

Looking forward...

ITer2022:
> > Suppose there are two servers.
> > 
> > 1. We set one to HYBRID or DISPATCH_ONLY mode and the other to GAME_ONLY mode. The function getAccountDatastore() returns gameDatastore on HYBRID-mode server, but dispatchDatasotre on GAME_ONLY-mode server, witch will cause white door error, cuz GAME_ONLY-mode server can not fetch account info.
> > 2. Now, we set both to HYBRID mode. The function getAccountDatastore() returns the same gameDatastore. We enter the same region actually even though we could choose regions on client.
> 
> That is an issue, I will look into it but you have solved the solution but in the wrong way which will cause issues in the future.

I think i've got your idea. Do u mean that the GAME_ONLY-mode server shouldn't connect to account db directly? It should fetch a login token dispatched by dispatch server instead considering secure reasons?

ITer2022:
> > Suppose there are two servers.
> > 
> > 1. We set one to HYBRID or DISPATCH_ONLY mode and the other to GAME_ONLY mode. The function getAccountDatastore() returns gameDatastore on HYBRID-mode server, but dispatchDatasotre on GAME_ONLY-mode server, witch will cause white door error, cuz GAME_ONLY-mode server can not fetch account info.
> > 2. Now, we set both to HYBRID mode. The function getAccountDatastore() returns the same gameDatastore. We enter the same region actually even though we could choose regions on client.
> 
> That is an issue, I will look into it but you have solved the solution but in the wrong way which will cause issues in the future.

I think i've got your idea. Do u mean that the GAME_ONLY-mode server shouldn't connect to account db directly? It should fetch a login token dispatched by dispatch server instead considering secure reasons?

exzork:
My previous idea was #1251 , will rewriting that on socket.io when i have spare time.

ITer2022:
> My previous idea was #1251 , will rewriting that on socket.io when i have spare time.

So, the bug does exist...

4Benj:
> > My previous idea was #1251 , will rewriting that on socket.io when i have spare time.
> 
> So, the bug does exist...

What @exzork is talking about is the previously mentioned plan to completely separate servers running in DISPATCH and GAME and have them communicate via websockets.

> I think i've got your idea. Do u mean that the GAME_ONLY-mode server shouldn't connect to account db directly? It should fetch a login token dispatched by dispatch server instead considering secure reasons?

Yes, it shouldn't but as of now it does as it is currently not implemented as said by @exzork, he already made a PR to implement this but he will be rewriting it in socket.io 

> Sorry, too. I didn't make it clear. I use the same UID to indicate that the players we created are the same between regions. Suppose we've already created a player in region1, when we tried to login to region2, we got the same player we created in region1, not a newly created.

This is possibly due to the region2 using the wrong datastore so it's fetching the player's data from region1 instead. However, I looked into some of the code and to me RegionHandler seems totally broken in the development branch and I cannot seem to connect to a second region which may be because I'm running the 2.8 branch which has changed how regions communicated with the game and not the RegionHandler issues that are also present in the development branch. I'm not sure if we should fix the current issues with the RegionHandler at the moment or wait for @exzork to implement the new solution.

This PR seems to attempt to fix something that shouldn't be an issue within the DatabaseHandler but somewhere else in the code and might cause issues in the future for existing or new servers who haven't adapted the new changes because the two datastores weren't used and shouldn't be needed when the server isn't in GAME_ONLY mode with the current solution.
I may look into this a bit more but I will be closing this issue for now since it seems to have found the wrong solution to the problem present and the solution may lie somewhere else in the code or because of a misconfigured server.

Feel free to look into the issue more to find the right solution but may be a waste of time because of the coming changes with separating the server types.

PR #1489, opened 2022-07-11
feature(support 2.8): upgrade grasscutter to 2.8

Konoyoru435:
## Description

***The patch is still under discussion and this pr may be replaced by backroom at any time***

Support Anime Game 2.8.

1. Upload new protos, and I added proto for almost all activities to help developers implement more activities. To solve the memory issues, add `gradle.properties`;
2. Fix some bugs cuz by new protos, a painful experience that includes a lot of guesswork;
3. Modified the login logic, from now on if you want to use the grasscutter for this branch you have to patch your `UserAssembly.dll`, but keep the seed exchange logic.

Because of the obfuscation, a large number of proto field names have been changed, and change to the original de-obfuscated logic and fields is unavoidable, so I can only rely on my personal guesses and related logical to fix them. I apologize if the functionality is broken.

BREAKING CHANGE: ALL

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation
- [x] New Version

## TODO List

- [ ] Remove useless proto;
- [ ] Provide comments for obfuscated fields.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] My pull request is unique and no other pull requests have been opened for these changes
- [x] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [x] I am responsible for any copyright issues with my code if it occurs in the future.

PR #1497, opened 2022-07-12
2.8 Support (Backrooms Edition)

4Benj:
## Description
This pull request brings Grasscutter to 2.8!
The main differences between this one and #1489, is that we've been developing this during the beta and that therefore it contains fixes that the other PR didn't have such as fixing the chat proto and implementing some scene id stuff for the new gaa, additionally that PR required patching the UA, and we patch the metadata instead which can open more (and secure) possibilities. 

Update: Thanks to @Yazawazi, this PR now supports UA patching. However, someone who patched via UA cannot join a server that has ``uaPatchCompatible`` disabled in the config, similarly someone who patched via metadata cannot join a server with ``uaPatchCompatible`` enabled in the config. A UA patched client and a metadata patched client cannot join the same server. Hopefully in the future, will we come up with a solution to connect the two differently patched clients, but for now, this is all we got... 

Additional Feature: Real Passwords!
This allows people to use passwords during login. However, this feature is marked as "experimental" and needs to be enabled in the config because it is currently incompatible with existing accounts for public servers due to the fact that currently, to reset your password you need to request it to be reset it from an administrator.

**This PR will require client patching! To make life easier there is a [draft PR on the Cultivation repo ](https://github.com/Grasscutters/Cultivation/pull/22) which can automatically patch and restore the game's metadata so you can use one client to play on both GC servers and official servers. (However, I recommend you only use Cultivation from now on).**

## Issues fixed by this PR
- Chat fixes
- Dispatch key exchange
- Fixes for GAA
among some other things

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


nitrog0d:
chat fixes my favorite fix tbh now I can finally see what I say on chat

Yazawazi:
The PR is now compatible with the UA patch client, you can enable the `uaPatchCompatible` option in the configuration. However, it is currently not compatible with the metadata patch and when you enable this option, and this compatibility option does not take into account password login.

Update: From now on, UA Patch players can also log in with their passwords normally!

MoonLGH (approved):
very real, and most important thing it work

KingRainbow44 (approved):
code looks good, functionality should be tested by write access

4Benj:
Alright now this PR is more reasonable lmao. Thanks @Birdulon for the clean up

PR #1512, opened 2022-07-14
Major quest system revision, new QuestTrigger handlers, miscellaneous bug fixes

akatatsu27:
## Description
-ContentTriggerFire will work if we switch lua library to jnlua
-QuestManager can now pass accept, fail and finish QuestTriggers to all active quests
-GameMainQuest is now responsible for handling its subQuests' conditions appropriately
-track and store in the database quest related talks in GameMainQuest (important for some conditions)
-Implement QuestGlobalVars
-Added various QuestTrigger handlers
-Initialize all GameMainQuest questVars to 0 (official server behaviour) 
-Handle QuestUpdateQuestVarReq/Rsp
-added start(), fail() and stub rewind() methods to GameQuest
-fix finishCond set to null if no acceptCond
-fix if accept/fail/finish CondComb is null, getAccept/Fail/FinishCondComb returns null (should return LOGIC_NONE)
-fix some packet handlers
-fix QuestManager::addQuest() notify player only one new subQuest added (there could be multiple)
-Made more clear in the code if referring to MainQuest or SubQuest
-Added comments to some wrongly handled QuestTriggers

I suggest testing by starting quest 100301 via console command
## Type of changes
- [x] Bug fix
- [ ] New feature 
- [x] Enhancement
- [x] Documentation

## Checklist:


ghost (approved):

src/main/java/emu/grasscutter/game/quest/handlers/QuestBaseHandler.java
```diff
@@ -1,10 +1,23 @@
...
 import emu.grasscutter.data.excels.QuestData.QuestCondition;
+import emu.grasscutter.game.player.Player;
+import emu.grasscutter.game.quest.GameMainQuest;
 import emu.grasscutter.game.quest.GameQuest;
+import jdk.jshell.spi.ExecutionControl;
 
 public abstract class QuestBaseHandler {
 
-	public abstract boolean execute(GameQuest quest, QuestCondition condition, String paramStr, int... params);
+	public  boolean execute(GameQuest quest, QuestCondition condition, String paramStr, int... params) throws ExecutionControl.NotImplementedException {
```
Akka0: why it divided into 3 methods here? you can access the main quest by `quest.getMainQuest()` and the player by `quest.getMainQuest().getOwner()`

akatatsu27: WIth how GameMainQuest::evaluateAccept/Fail/FinishCond is called, unless subQuest is in params, we can't determine which subQuest. Also, in many QuestTriggers, which subQuest is not relevant. Finally, I think this improves code readability.

Akka0: i think there is no need to determine which subquest here, because a loop will check all of the subquests in main quests and auto accept the quests which meet the condition. And it cause hard to maintain by repeating too much code in `ServerQuestHandler` and an unnecessarily `switch` in `evaluateAcceptCond` etc. 
ideally, we can just add a Handler to support the new trigger, no need to add a `case`

src/main/java/emu/grasscutter/game/quest/GameMainQuest.java
```diff
@@ -120,9 +124,233 @@ public void finish() {
...
+            ArrayList<QuestData> subQuests = new ArrayList<>();
+            for(SubQuestData subQuestData: subQuestsData) {
+                subQuests.add(GameData.getQuestDataMap().get(subQuestData.getSubId()));
+            }
+            /*
+                Make sure we check the subQuests with the exact condition AND parameters
+                Note: because we currently don't pass all parameters (example: QUEST_CONTENT_ENTER_DUNGEON normally has 2 parameters),
+                I'll only check for params[0], which is a worse filter
+            */
+            List<QuestData> subQuestsWithCond = subQuests.stream()
```
Akka0: also, i think it is unnecessary. when it comes to a condition which is not meet or not impl, the trigger will just return `false` so that the quest will not be accepted.
`q.getParam()[0] == params[0]` this

akatatsu27: I agree. but some QuestTriggers (example: QUEST_CONTENT_GAME_TIME_TICK) have same params ,but different paramStr.
I'll make it `q.getParam()[0] == params[0] && q.getParamStr.contains(paramStr)`

src/main/java/emu/grasscutter/game/quest/handlers/QuestBaseHandler.java
akatatsu27: ok I see. I'll revert this, and move the debug messages to the appropriate handlers

akatatsu27: I agree for fail and finish, but for evaluateAcceptCond, quest is always `null`. We would have to add all subQuests and set state to QUEST_STATE_UNSTARTED. I'll see what I can do. Thanks for the feedback!

Akka0: for accept, it should check the quest metadata; for fail or finish, it should check the quest which player holds it

src/main/java/emu/grasscutter/game/quest/GameMainQuest.java
Akka0: i think it is hardcode. if they have the same params, they should all get triggered. It wont cause any exceptions when more than one quest progress change at the same time
And we will do the comparison of `params` and `paramStr` at the trigger, so it is unnecessary here

akatatsu27: Looking at QuestExcelConfigData again, there area lot of different conditions with `params` = [0,0]. I think triggering many handlers that we know are going to return false is going to scale very poorly with a larger playerbase, who each  might have several quests active. Isn't filtering beforehand more computationally efficient? maybe we should check `condition` and `params` and `paramStr`

src/main/java/emu/grasscutter/game/quest/handlers/QuestBaseHandler.java
akatatsu27: we'd still need to have a different QuestBaseHandler::execute for accepting quests though, so `ServerQuestHandler` would still be messy. I think initializing all subQuests to `QUEST_STATE_UNSTARTED` is cleaner, and it's probably what Certain Anime Game Company is doing. When state = `PARENT_QUEST_STATE_FINISHED` or `PARENT_QUEST_STATE_FAILED` or `PARENT_QUEST_STATE_CANCELED`, we could remove all the subQuests from the MainQuest  from the database to save space afterwards

src/main/java/emu/grasscutter/game/quest/GameMainQuest.java
Akka0: but the problem is the conditions are not always `==`, like `QUEST_COND_PLAYER_LEVEL_EQUAL_GREATER` it is `>=`, the pre-check will make it broken

akatatsu27: indeed... also in `QUEST_CONTENT_GAME_TIME_TICK` we are given a range. It seems without adding special cases, the simplest filter is ` q.getType() == condType`

tnsc4502:
@akatatsu27 
Hi, I often encounter the error like below when I manually add quest (for example, /quest add 37001):

`java.lang.NullPointerException: Cannot invoke "emu.grasscutter.game.quest.ServerQuestHandler.triggerCondition(emu.grasscutter.game.quest.GameQuest, emu.grasscutter.data.excels.QuestData$QuestCondition, String, int[])" because the return value of "emu.grasscutter.game.quest.GameMainQuest.getServerQuestHandler()" is null
        at emu.grasscutter.game.quest.GameMainQuest.tryAcceptSubQuests(GameMainQuest.java:158)
        at emu.grasscutter.game.quest.QuestManager.triggerEvent(QuestManager.java:233)
        at emu.grasscutter.game.quest.QuestManager.triggerEvent(QuestManager.java:210)
        at emu.grasscutter.game.quest.GameQuest.start(GameQuest.java:97)
        at emu.grasscutter.game.quest.QuestManager.addQuest(QuestManager.java:185)`

Also, sometimes the quest also failed to execute correctly due to the null data of triggerData, I think it is the exact same problem as aforementioned.

There is also another problem in my merged version, for instance, at the bottom of GameQuest.start(), the line:

`this.getOwner().getQuestManager().triggerEvent(QuestTrigger.QUEST_CONTENT_QUEST_STATE_EQUAL, this.getSubQuestId(), this.getState().getValue(),0,0,0);`

produces incorrect params that will be used, for example, in ContentQuestVarEqual which uses param[0] as an index to check quest var, however in this case param[0] seems to be this.getSubQuestId() that obviously is out of the range.

Maybe all the problems are caused by some failure during auto merging, if anyone is interested, you can checkout my branch, which is also based on jnlua version: https://github.com/tnsc4502/Grasscutter/tree/jnlua_v5.2

akatatsu27:
> @akatatsu27 Hi, I often encounter the error like below when I manually add quest (for example, /quest add 37001):
> 
> `java.lang.NullPointerException: Cannot invoke "emu.grasscutter.game.quest.ServerQuestHandler.triggerCondition(emu.grasscutter.game.quest.GameQuest, emu.grasscutter.data.excels.QuestData$QuestCondition, String, int[])" because the return value of "emu.grasscutter.game.quest.GameMainQuest.getServerQuestHandler()" is null at emu.grasscutter.game.quest.GameMainQuest.tryAcceptSubQuests(GameMainQuest.java:158) at emu.grasscutter.game.quest.QuestManager.triggerEvent(QuestManager.java:233) at emu.grasscutter.game.quest.QuestManager.triggerEvent(QuestManager.java:210) at emu.grasscutter.game.quest.GameQuest.start(GameQuest.java:97) at emu.grasscutter.game.quest.QuestManager.addQuest(QuestManager.java:185)`
> 
> Also, sometimes the quest also failed to execute correctly due to the null data of triggerData, I think it is the exact same problem as aforementioned.
> 
> There is also another problem in my merged version, for instance, at the bottom of GameQuest.start(), the line:
> 
> `this.getOwner().getQuestManager().triggerEvent(QuestTrigger.QUEST_CONTENT_QUEST_STATE_EQUAL, this.getSubQuestId(), this.getState().getValue(),0,0,0);`
> 
> produces incorrect params that will be used, for example, in ContentQuestVarEqual which uses param[0] as an index to check quest var, however in this case param[0] seems to be this.getSubQuestId() that obviously is out of the range.
> 
> Maybe all the problems are caused by some failure during auto merging, if anyone is interested, you can checkout my branch, which is also based on jnlua version: https://github.com/tnsc4502/Grasscutter/tree/jnlua_v5.2

I was about to convert to draft to fix the aforementioned and some other related issues. I'm afraid I'm unfamiliar with jnlua so your help would be very appreciated! Is there somewhere I can contact you? my name is the same on discord btw

akatatsu27:
Recent big changes in the development branch caused my branch to have severe conflicts. Will make a new PR

PR #1529, opened 2022-07-18
Refactor Command usage and description strings

### LINKED ISSUE #1519: [Bug] Cannot scroll through help command

**Describe the bug**

The help command is kind of useless now that it's sent all as one message because it cannot be scrolled. The command used to work better when it sent each command as its own individual message, however now it's impossible to scroll through the output of the command. As seen in the video below, it is impossible to get any more information than what's at the bottom.

**Which branch did you use?**
Grasscutter version: 1.2.2-dev
Git Hash: e7fb5362

**Screenshots**

https://user-images.githubusercontent.com/104761702/179389329-24af7756-7de3-4006-ac6e-fe08838a0375.mp4

**Additional context**

The old system sent it as many individual messages. I'm not sure of exactly when things started breaking.

Birdulon:
## Description
Changes `usage` string to only include arguments, and automatically generate usage strings from the command label or shortest alias as well as the per-language "Usage: " prefix, and the targeting information (which also handles required vs. optional for console). Additionally makes `usage` an array of strings instead of just one, so that this is possible:
image

Removes `description` and replaces usage of it with a new method which fetches the translation key of the form `commands.{label}.description`.

Attempts to make the annotation declaration order on `CommandHandler`s consistently match the `Command` annotation declaration order (I may have missed some, please don't let me merge without fixing them!)

Makes `help` send one message per command instead of all in one message.

Attempts to move the usage syntax more towards the standard form outlined at https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key
- `[]` blocks denote optional args. `[a] [b] [c]` means `a`, `b c`, and `` are all valid.
- - If optional args depend on previous optional args, they must be written like `[this [maybe this [and this]]]`
- Anything inside `<>` is a name that must be replaced with a value. e.g. `lv<level>` means you write `lv10`.
- - Conversely, anything outside of `<>` is literal: `[level]` means you write `level`, not `10`!
- `(a|b|c)` is the notation for choose-one. Previously we did things like `<a|b|c>` or `[a|b|c]`.
- - `<enemyId|entityId>` would be `(<enemyId>|<entityId>)`
- - `[enemyId|entityId]` would be `[(<enemyId>|<entityId>)]`
- - Due to deficiencies in the previous format, we had a few instances of `<thing|"literal">`, which would now be `(<thing>|literal)`
I've likely missed some of these replacements, watch this space.

## Issues fixed by this PR
Fixes #1519.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


ghost (approved):

lunaticwhat:
nice job <3

PR #1557, opened 2022-07-23
Read Open States from Excels

### LINKED ISSUE #1535: [Bug] OpenStates not working since 1fe304c - cannot log into my account

Yes.

**Describe the bug**
This has been happening ever since 1fe304c:
image
Any earlier version will work and current versions do not. I'm reporting this as a bug because another person in the discord server is reporting the same problem and I have not done anything like messing with the source code or database to cause this inexplicable issue.. The error does mention openstates, which is weird because 051198c deals with this while 1fe304c doesn't, however that is still the commit where this account first started suffering these problems. I did try 949916a and there was nothing wrong. If needed, I will provide my player data, however messing with said data (i.e. changing my position) results in the same issue.

**Which branch did you use?**
Grasscutter version: 1.2.2-dev
Git Hash: 1fe304c

**Screenshots**

**Additional context**

I doubt that this is anything to do with my position in the game world - the last place I logged out before updating was next to a domain in Mondstadt.

GanyusLeftHorn:
## Description

This PR removes the hardcoded `OpenState.java` and instead reads them from `OpenStateConfigData.json`. It also makes some refactorings for `PlayerOpenStateManager` and adds subcommands to `/setprop`:

- Open states are now read from `OpenStateConfigData.json`, including information on how they are unlocked.
- The set of open states that are unlocked on every account by default now comprises:
    * All states that are unlocked by default according to the config (`defaultState = true`).
    * `OPEN_STATE_PAIMON` (1), which we need to unlock to give the player access to a functioning chat for commands.
    * All states that have unlock conditions that we don't handle yet (quests, parent quests, reputation level etc.). This could (and should) be changed to NOT include states which are set by the client (`allowClientOpen = true`), however, there is a problem when the client shows UI hints and forces the player to click on UI icons that don't yet work correctly. In such a case, the hint on screen will never disappear. We could solve this by hardcoding the necessary states, like `OPEN_STATE_PAIMON` (or by making all the necessary UI buttons work ...).
    * NOT any of the states listed in `BLACKLIST_OPEN_STATES`. This is currently used for the border around the world map. As this is a quest-unlocked state (using some high-id dummy quest), this can be removed as soon as quest unlocks are fully implemented.
- The `/setprop` command was extended to allow temporarily setting and unsetting of open states for development and testing perposes. `/setprop setopenstate X` sets open state X, and `/setprop unsetopenstate X` removes the state. This is not saved to the DB, so the effects of this are purely temporary.

## Issues fixed by this PR

Fixes #1535 

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


src/main/java/emu/grasscutter/data/excels/OpenStateData.java
```diff
@@ -1,13 +1,58 @@
...
+    public static String PARENT_QUEST_UNLOCK_COND = "OPEN_STATE_COND_PARENT_QUEST";
+
+    private int id;
+    @Getter private boolean defaultState;
+    @Getter private boolean allowClientOpen;
+    @Getter private int systemOpenUiId;
+    @Getter private List<OpenStateCond> cond;
+
+    public static class OpenStateCond {
+        @Getter private String condType;
```
ghost: `condType` should be an enum instead of String. Gson automatically parses them to an enum.

ghost (changes requested):

src/main/java/emu/grasscutter/data/excels/OpenStateData.java
GanyusLeftHorn: Changed.

ghost (approved):

PR #1574, opened 2022-07-26
Add QuestEncryptionKeys.json

TotallyNotOndre:
## Description
#1561 forgot to add [QuestEncryptionKeys.json](https://gist.github.com/Hartie95/8b035f144fa17d5c43d17665ee460ba7#file-questencryptionkeys-json) into a data folder so i leave it here

## Issues fixed by this PR
- Fixes `<ERROR:emu.grasscutter.data.ResourceLoader> Unable to load quest keys.`
Before fix:
Before
After fix:
After

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


ghost:
Cutscene encryption keys should be put in the `resources` folder since they are extracted from the client. (Yes, i know theyre currently loaded from the `data` folder)

Hartie95:
It should also not break anything if the file is not there. The file is also only needed for 2.7+ prerendered cutscene support.

Hartie95:
> Cutscene encryption keys should be put in the `resources` folder since they are extracted from the client. (Yes, i know theyre currently loaded from the `data` folder)

They are not really extracted from the client, for 2.7+ they aren't included in the client and get only send by the server as far as I know.

ghost:
> They are not really extracted from the client, for 2.7+ they aren't included in the client and get only send by the server as far as I know.

That is true, I still think they blong more in resources than data though.

PR #1591, opened 2022-07-27
Toggle constellation

juuyokka:
## Description

Allows `setConst` command to toggle individual constellations, enabling unofficial combinations of constellations. Further discussion is needed on what `coreProudSkillLevel` should be set to when these unintended constellation states are in effect.

Constellation screen preview
Talent screen preview

## Issues fixed by this PR

Fixed `manage_languages.py` to read language files as `UTF-8` instead of the default encoding

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


juuyokka:
To ensure that `coreProudSkillLevel` is well defined, the function `setCoreProudSkillLevel` has been removed and `getCoreProudSkillLevel` will calculate the constellation level on the fly instead of retrieving it as a field in the `Avatar` class.

src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
```diff
@@ -14,72 +14,109 @@
...
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 @Command(
     label = "setConst",
-    aliases = {"setconstellation"},
-    usage = {"<constellation level>"},
+    aliases = {"setconstellation", "setcons", "constellations", "cons"},
+    usage = {"<constellation level>", "[(set|toggle)] <constellation level>"},
```
Birdulon:
```suggestion
    usage = {"[set] <constellation level>", "toggle <constellation level>"},
```

src/main/java/emu/grasscutter/server/packet/recv/HandlerUnlockAvatarTalentReq.java
```diff
@@ -14,7 +14,7 @@ public void handle(GameSession session, byte[] header, byte[] payload) throws Ex
         UnlockAvatarTalentReq req = UnlockAvatarTalentReq.parseFrom(payload);
 
         // Unlock avatar const
-        session.getServer().getInventorySystem().unlockAvatarConstellation(session.getPlayer(), req.getAvatarGuid());
+        session.getServer().getInventorySystem().unlockNextAvatarConstellation(session.getPlayer(), req.getAvatarGuid());
```
Birdulon: Might be worth actually using `req.getTalentId()` to correctly handle the very-non-vanilla case where the player can actually unlock different constellations than the lowest locked one.
image

src/main/java/emu/grasscutter/game/avatar/Avatar.java
```diff
@@ -432,11 +427,13 @@ public Set<Integer> getTalentIdList() {
     }
 
     public int getCoreProudSkillLevel() {
-        return coreProudSkillLevel;
```
Birdulon: Should be safe to remove the `Avatar.coreProudSkillLevel` property now, since it's no longer used anywhere.

Birdulon (approved):
Tested f15a7e7 and confirmed that command usage and general constellation unlocking work as expected.

src/main/java/emu/grasscutter/command/commands/ResetConstCommand.java
```diff
@@ -38,7 +38,6 @@ public void execute(Player sender, Player targetPlayer, List<String> args) {
 
     private void resetConstellation(Avatar avatar) {
         avatar.getTalentIdList().clear();
-        avatar.setCoreProudSkillLevel(0);
```
juuyokka: `resetConstCommand` essentially does the same thing as `setConstCommand`, but with less features. Perhaps this command can be removed entirely.

Birdulon: I agree.

juuyokka: I could also add a `/cons reset` subcommand that would internally call `/cons set 0` if we want to.

Birdulon: Only other regression would be "reset all"

juuyokka:
`setConstCommand` now has a `reset` subcommand. I have only tested it in singleplayer.

Birdulon:
# Recent DMCA compliance push closed active PRs.

This particular PR would need a little bit of changing to match up to recent constellation changes (should be easier to implement the toggle now)

PR #1640, opened 2022-08-10
New translate features

Birdulon:
## Description
All arguments to `translate` will now be converted to strings. If a `TextStrings` is passed (i.e. from the textmap lookups), it will pick the correct localised string from within.

Cleans up commands that were using `Integer::toString` and `Float::toString` for `translate` arguments.

Adds talent name and desc feedback to the talent command.
image
image

Will probably make some other tweaks to the language backend. Might even get around to handbook search.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


KingRainbow44:
a component system (similar to how [Paper](https://papermc.io) does it) might be nice with grasscutter (only maybe, would be hell to convert everything to it) as colors in-game and in the server console are represented differently, though this should probably be another PR.

Birdulon:
We can probably settle for stripping xml tags when printing to console

Issue #1725, opened 2022-08-29
[Bug] 


w3bOs:
**Did you look for other closed issues that have the same problem?**

Yes
**Describe the bug**

Appears to be a scene error i think

**Which branch did you use?**

**Screenshots**
08:27:51 <ERROR:SceneScriptManager> [LUA] call trigger failed condition_EVENT_ENTER_REGION_289,emu.grasscutter.scripts.data.ScriptArgs@5856b839
org.luaj.vm2.LuaError: script:2394 bad argument: userdata expected, got number
        at org.luaj.vm2.LuaValue.argerror(Unknown Source)
        at org.luaj.vm2.LuaValue.checkuserdata(Unknown Source)
        at org.luaj.vm2.lib.jse.JavaMethod.call(Unknown Source)
        at org.luaj.vm2.LuaClosure.execute(Unknown Source)
        at org.luaj.vm2.LuaClosure.call(Unknown Source)
        at emu.grasscutter.scripts.SceneScriptManager.safetyCall(SceneScriptManager.java:375)
        at emu.grasscutter.scripts.SceneScriptManager.callScriptFunc(SceneScriptManager.java:368)
        at emu.grasscutter.scripts.SceneScriptManager.realCallEvent(SceneScriptManager.java:323)
        at emu.grasscutter.scripts.SceneScriptManager.lambda$callEvent$10(SceneScriptManager.java:308)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:833)

**Additional context**

Is a weird bug. If it appears, some chests dissapear or something.

akatatsu27:
duplicate of #1703

Hartie95:
As I mentioned here: https://github.com/Grasscutters/Grasscutter/issues/1703#issuecomment-1239836750

[...]

Issue #1703, opened 2022-08-24
[Bug] Introduction (tutorial?) quests seem to throw errors about Lua


SoftwareGuy:
**Did you look for other closed issues that have the same problem?**

Yes, but there weren't any.

**Describe the bug**

When trying to complete the beginning quests, the server logs some errors in the console about it expecting a `userdata` value type but instead it got a number. Refer to additional context below for how to replicate.

```
22:39:41 <ERROR:SceneScriptManager> [LUA] call trigger failed condition_EVENT_ENTER_REGION_24,emu.grasscutter.scripts.data.ScriptArgs@2038d7b2
org.luaj.vm2.LuaError: script:967 bad argument: userdata expected, got number
        at org.luaj.vm2.LuaValue.argerror(Unknown Source)
        at org.luaj.vm2.LuaValue.checkuserdata(Unknown Source)
        at org.luaj.vm2.lib.jse.JavaMethod.call(Unknown Source)
        at org.luaj.vm2.LuaClosure.execute(Unknown Source)
        at org.luaj.vm2.LuaClosure.call(Unknown Source)
        at emu.grasscutter.scripts.SceneScriptManager.safetyCall(SceneScriptManager.java:374)
        at emu.grasscutter.scripts.SceneScriptManager.callScriptFunc(SceneScriptManager.java:367)
        at emu.grasscutter.scripts.SceneScriptManager.realCallEvent(SceneScriptManager.java:322)
        at emu.grasscutter.scripts.SceneScriptManager.lambda$callEvent$10(SceneScriptManager.java:308)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:833)
22:39:42 <INFO:Scene> Scene 3 loaded 2 group(s)
22:39:44 <INFO:Scene> Scene 3 loaded 2 group(s)
> quest finish 35200
22:41:12 <INFO:CommandHandler> Finished quest 35200.
[...]

PR #1756, opened 2022-09-07
fix ScriptLib (aggressive)

pfyy:
## Description

This is an **aggressive** fix for ScriptLib. Since luaj is not likely to be replaced and has numerous bug-like behaviors, this fix attempts to modify the GC and **its resources** to resolve these problems.

The modified resources can be found here:
[https://github.com/pfyy/Grasscutter_Resources](https://github.com/pfyy/Grasscutter_Resources)

Only the scripts are modified. For every `ScriptLib.foo()` call, an argument `ScriptLib` is inserted at the beginning, e.g. `ScriptLib.foo(bar)` becomes `ScriptLib.foo(ScriptLib, bar)`.

## Issues fixed by this PR

#1703
#1725

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/scripts/ScriptLib.java
+++ b/src/main/java/emu/grasscutter/scripts/ScriptLib.java
@@ -62,7 +62,7 @@ public class ScriptLib {
 	public void removeCurrentGroup(){
 		this.currentGroup.remove();
 	}
-	public int SetGadgetStateByConfigId(int configId, int gadgetState) {
+	public int SetGadgetStateByConfigId(Object context, int configId, int gadgetState) {
 		logger.debug("[LUA] Call SetGadgetStateByConfigId with {},{}",
 				configId,gadgetState);
 		Optional<GameEntity> entity = getSceneScriptManager().getScene().getEntities().values().stream()
@@ -80,7 +80,7 @@ public class ScriptLib {
 		return 1;
 	}
 
-	public int SetGroupGadgetStateByConfigId(int groupId, int configId, int gadgetState) {
+	public int SetGroupGadgetStateByConfigId(Object context, int groupId, int configId, int gadgetState) {
 		logger.debug("[LUA] Call SetGroupGadgetStateByConfigId with {},{},{}",
 				groupId,configId,gadgetState);
 
@@ -93,7 +93,7 @@ public class ScriptLib {
 		return 0;
 	}
 	
-	public int SetWorktopOptionsByGroupId(int groupId, int configId, int[] options) {
+	public int SetWorktopOptionsByGroupId(Object context, int groupId, int configId, int[] options) {
 		logger.debug("[LUA] Call SetWorktopOptionsByGroupId with {},{},{}",
 				groupId,configId,options);
 		
@@ -115,12 +115,12 @@ public class ScriptLib {
 		return 0;
 	}
 
-	public int SetWorktopOptions(LuaTable table){
+	public int SetWorktopOptions(Object context, LuaTable table){
 		logger.debug("[LUA] Call SetWorktopOptions with {}", printTable(table));
 		// TODO
 		return 0;
 	}
-	public int DelWorktopOptionByGroupId(int groupId, int configId, int option) {
+	public int DelWorktopOptionByGroupId(Object context, int groupId, int configId, int option) {
 		logger.debug("[LUA] Call DelWorktopOptionByGroupId with {},{},{}",groupId,configId,option);
 
 		Optional<GameEntity> entity = getSceneScriptManager().getScene().getEntities().values().stream()
@@ -141,7 +141,7 @@ public class ScriptLib {
 	}
 	
 	// Some fields are guessed
-	public int AutoMonsterTide(int challengeIndex, int groupId, Integer[] ordersConfigId, int tideCount, int sceneLimit, int param6) {
+	public int AutoMonsterTide(Object context, int challengeIndex, int groupId, Integer[] ordersConfigId, int tideCount, int sceneLimit, int param6) {
 		logger.debug("[LUA] Call AutoMonsterTide with {},{},{},{},{},{}",
 				challengeIndex,groupId,ordersConfigId,tideCount,sceneLimit,param6);
 
@@ -156,7 +156,7 @@ public class ScriptLib {
 		return 0;
 	}
 	
-	public int AddExtraGroupSuite(int groupId, int suite) {
+	public int AddExtraGroupSuite(Object context, int groupId, int suite) {
 		logger.debug("[LUA] Call AddExtraGroupSuite with {},{}",
 				groupId,suite);
 		SceneGroup group = getSceneScriptManager().getGroupById(groupId);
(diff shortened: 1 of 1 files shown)


pfyy:
The code for modifying the scripts is given below:
```python
import os
from luaparser import ast, astnodes


v: list[dict[str]] = []


def traverse(node: astnodes, src: list[str]):
    if isinstance(node, astnodes.Chunk):
        v.append(dict())
        traverse(node.body, src)
        v.pop()
    elif isinstance(node, astnodes.Block):
        v.append(dict())
        for i in node.body:
            traverse(i, src)
        v.pop()
    elif isinstance(node, astnodes.Statement):
        if isinstance(node, astnodes.Assign):
            if isinstance(node, astnodes.LocalAssign):
                if node.values:
                    for i, j in zip(node.targets, node.values):
                        name = i.id
                        value = traverse(j, src)
                        v[-1][name] = value
                else:
                    for i in node.targets:
                        name = i.id
                        v[-1][name] = None
            else:
                for i, j in zip(node.targets, node.values):
                    if isinstance(i, astnodes.Name):
                        name = i.id
                        value = traverse(j, src)
                        for k in range(len(v)-1, -1, -1):
                            if name in v[k]:
                                break
                        v[k][name] = value
        elif isinstance(node, astnodes.While):
            traverse(node.body, src)
        elif isinstance(node, astnodes.Do):
            traverse(node.body, src)
        elif isinstance(node, astnodes.Repeat):
            traverse(node.body, src)
        elif isinstance(node, astnodes.ElseIf):
            traverse(node.body, src)
        elif isinstance(node, astnodes.If):
            traverse(node.body, src)
        elif isinstance(node, astnodes.Fornum):
            traverse(node.body, src)
        elif isinstance(node, astnodes.Forin):
            traverse(node.body, src)
        elif isinstance(node, astnodes.Call):
            if isinstance(node.func, astnodes.Name):
                name = node.func.id
                for i in range(len(v)-1, -1, -1):
                    if name in v[i]:
                        if isinstance(v[i][name], tuple):
                            j = node.line-1
                            if src[j].find(f"{name}()") != -1:
                                src[j] = src[j].replace(
                                    f"{name}()", f"{name}(ScriptLib)"
                                )
                            else:
                                src[j] = src[j].replace(
                                    f"{name}(", f"{name}(ScriptLib, "
                                )
                        break
        elif isinstance(node, astnodes.Function):
            traverse(node.body, src)
        elif isinstance(node, astnodes.LocalFunction):
            traverse(node.body, src)
    elif isinstance(node, astnodes.Name):
        name = node.id
        if name == "ScriptLib":
            return "ScriptLib"
    elif isinstance(node, astnodes.Index):
        if isinstance(node.value, astnodes.Name):
            name = node.value.id
            for i in range(len(v)-1, -1, -1):
                if name in v[i]:
                    if v[i][name] == "ScriptLib":
                        return "ScriptLib", node.idx.id


mod_folder = "mod"


def process(path):
    with open(path) as f:
        src = f.readlines()

    tree = ast.parse(''.join(src))

    traverse(tree, src)

    new_path = os.path.join(mod_folder, path)

    os.makedirs(os.path.dirname(new_path), exist_ok=True)

    with open(new_path, "w") as f:
        f.writelines(src)


for i, _, j in os.walk("Scripts"):
    for k in j:
        process(os.path.join(i, k))

```

Hartie95:
Why do you add Scriptlib as a parameter? the current Resources with my latest changes+ normal methods without scriptlib parameter seem to work without a problem.
All you needed for most calls to work, was adding the LuaContext to the calls witout one, and everything works.
Some other problems are still in the Scriptlib, like getEntityById, which is not yet working, together with some missing methods, but it generally works with luaj with that.

pfyy:
> Why do you add Scriptlib as a parameter? the current Resources with my latest changes+ normal methods without scriptlib parameter seem to work without a problem. All you needed for most calls to work, was adding the LuaContext to the calls witout one, and everything works. Some other problems are still in the Scriptlib, like getEntityById, which is not yet working, together with some missing methods, but it generally works with luaj with that.

Sorry, I didn't notice that some similar work is already done.

The reason is that I find the implementation of `JavaMethod` is using `method.invoke(instance, a)`. But scripts cannot always guarantee that the first parameter is `instance`, e.g. `ScriptLib.GetEntityType`.

https://github.com/luaj/luaj/blob/master/src/jse/org/luaj/vm2/lib/jse/JavaMethod.java

```java
class JavaMethod extends JavaMember {

	static final Map methods = Collections.synchronizedMap(new HashMap());
	
	static JavaMethod forMethod(Method m) {
		JavaMethod j = (JavaMethod) methods.get(m);
		if ( j == null )
			methods.put( m, j = new JavaMethod(m) );
		return j;
	}
	
	static LuaFunction forMethods(JavaMethod[] m) {
		return new Overload(m);
	}
	
	final Method method;
	
	private JavaMethod(Method m) {
		super( m.getParameterTypes(), m.getModifiers() );
		this.method = m;
		try {
			if (!m.isAccessible())
				m.setAccessible(true);
		} catch (SecurityException s) {
		}
	}

	public LuaValue call() {
		return error("method cannot be called without instance");
	}

	public LuaValue call(LuaValue arg) {
		return invokeMethod(arg.checkuserdata(), LuaValue.NONE);
	}

	public LuaValue call(LuaValue arg1, LuaValue arg2) {
		return invokeMethod(arg1.checkuserdata(), arg2);
	}
	
	public LuaValue call(LuaValue arg1, LuaValue arg2, LuaValue arg3) {
		return invokeMethod(arg1.checkuserdata(), LuaValue.varargsOf(arg2, arg3));
	}
	
	public Varargs invoke(Varargs args) {
		return invokeMethod(args.checkuserdata(1), args.subargs(2));
	}
	
	LuaValue invokeMethod(Object instance, Varargs args) {
		Object[] a = convertArgs(args);
		try {
			return CoerceJavaToLua.coerce( method.invoke(instance, a) );
		} catch (InvocationTargetException e) {
			throw new LuaError(e.getTargetException());
		} catch (Exception e) {
			return LuaValue.error("coercion error "+e);
		}
	}
	
	/**
	 * LuaValue that represents an overloaded Java method.
	 * <p>
	 * On invocation, will pick the best method from the list, and invoke it.
	 * <p>
	 * This class is not used directly.  
	 * It is returned by calls to calls to {@link JavaInstance#get(LuaValue key)} 
	 * when an overloaded method is named.
	 */
	static class Overload extends LuaFunction {

		final JavaMethod[] methods;
		
		Overload(JavaMethod[] methods) {
			this.methods = methods;
		}

		public LuaValue call() {
			return error("method cannot be called without instance");
		}

		public LuaValue call(LuaValue arg) {
			return invokeBestMethod(arg.checkuserdata(), LuaValue.NONE);
		}

		public LuaValue call(LuaValue arg1, LuaValue arg2) {
			return invokeBestMethod(arg1.checkuserdata(), arg2);
		}
		
		public LuaValue call(LuaValue arg1, LuaValue arg2, LuaValue arg3) {
			return invokeBestMethod(arg1.checkuserdata(), LuaValue.varargsOf(arg2, arg3));
		}
		
		public Varargs invoke(Varargs args) {
			return invokeBestMethod(args.checkuserdata(1), args.subargs(2));
		}

		private LuaValue invokeBestMethod(Object instance, Varargs args) {
			JavaMethod best = null;
			int score = CoerceLuaToJava.SCORE_UNCOERCIBLE;
			for ( int i=0; i<methods.length; i++ ) {
				int s = methods[i].score(args);
				if ( s < score ) {
					score = s;
					best = methods[i];
					if ( score == 0 )
						break;
				}
			}
			
			// any match? 
			if ( best == null )
				LuaValue.error("no coercible public method");
			
			// invoke it
			return best.invokeMethod(instance, args);
		}
	}

}
```

PR #1762, opened 2022-09-08
Update zh-CN.json

2y8e9h22:
Translate Chinese briefly.


Birdulon:
Please don't reformat language files, it will get overwritten next lint push.

Birdulon (approved):

Issue #1476, opened 2022-07-09
[Bug] 攻击部分可交互实体会使其消失


z7087:
**Did you look for other closed issues that have the same problem?**

  https://github.com/Grasscutters/Grasscutter/issues/1405

**Describe the bug**

  当我持续攻击宝箱或部分可被攻击到的实体时，实体一段时间后会消失。

**Which branch did you use?**

  https://github.com/Grasscutters/Grasscutter/actions/runs/2641261047
  Grasscutter version: 1.2.2-dev
  Git Hash: 0141dce

**Screenshots**

![11111](https://user-images.githubusercontent.com/75190047/178110848-82154153-6fdb-4916-a511-daafb4222521.png)
![22222](https://user-images.githubusercontent.com/75190047/178110961-076854b9-80eb-4dca-855a-48db6c1dcbc9.png)

**Additional context**

wumulong-469:
![Screenshot_2022-07-10-07-08-46-277_com.miHoYo.GenshinImpact.jpg](https://user-images.githubusercontent.com/57058163/178125723-cc702d30-4e24-4189-ac5f-8be06fb59365.jpg)

![Screenshot_2022-07-10-07-08-59-617_com.miHoYo.GenshinImpact.jpg](https://user-images.githubusercontent.com/57058163/178125728-cb243dd3-0db9-4fc1-809e-4181ac62758a.jpg)

我也发现了，机关可以打掉。矿好像也遵循这个道理，用单手剑或者长枪挖矿没掉落，在黄条还没打完直接消失。

zhaodice:
已修复 https://github.com/Grasscutters/Grasscutter/pull/1606

Issue #1405, opened 2022-06-30
宝箱被攻击会消失[Bug] 


lniuua:
**Did you look for other closed issues that have the same problem?**

**Describe the bug**

**Which branch did you use?**

**Screenshots**

**Additional context**

xNissX233:
I have the same issue. Chests will disappear when hit by an attack or skill. They just vanish and don't drop anything.

Liujunhua007:
为什么我用action最新那个没宝箱...怎么开启呀

lniuua:
攻击某些树木和可以互动的实体也会消失

KiminaNatsuki:
Please follow the issue format, then feel free to make another issue.
拜託請先按照Issue格式提出問題，隨後再重發一次。

PR #1773, opened 2022-09-13
fixed gadget hp properties and invincibility handling

Hartie95:
## Description
Some gadgets, including chests, didn't have fightingProperties for hp, so they would just disapear when hit by an attack.
This merge requests reads the ConfigGadget definitions from from the json files in `/BinOutput/Gadget/` and applies the `HP`, `isLockHP` and `isInvincible` properties from combat.properties of those json objects.
The config to use is defined by GadgetData.jsonName.

## Issues fixed by this PR
#1476
#1405

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/data/binout/ConfigGadgetCombatProperty.java
+++ b/src/main/java/emu/grasscutter/data/binout/ConfigGadgetCombatProperty.java
@@ -1,0 +1,19 @@
+package emu.grasscutter.data.binout;
+
+import com.google.gson.annotations.SerializedName;
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.experimental.FieldDefaults;
+
+@Data
+@FieldDefaults(level = AccessLevel.PRIVATE)
+public class ConfigGadgetCombatProperty {
+    float HP;
+    boolean isLockHP;
+    boolean isInvincible;
+    boolean isGhostToAllied;
+    float attack;
+    float defence;
+    float weight;
+    boolean useCreatorProperty;
+}
--- a/src/main/java/emu/grasscutter/game/entity/GameEntity.java
+++ b/src/main/java/emu/grasscutter/game/entity/GameEntity.java
@@ -106,6 +106,10 @@ public abstract class GameEntity {
         return this.getFightProperties().getOrDefault(prop.getId(), 0f);
     }
 
+    public boolean hasFightProperty(FightProperty prop) {
+        return this.getFightProperties().containsKey(prop.getId());
+    }
+
     public void addAllFightPropsToEntityInfo(SceneEntityInfo.Builder entityInfo) {
         for (Int2FloatMap.Entry entry : this.getFightProperties().int2FloatEntrySet()) {
             if (entry.getIntKey() == 0) {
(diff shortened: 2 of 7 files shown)


src/main/java/emu/grasscutter/game/entity/GameEntity.java
```diff
@@ -160,8 +164,8 @@ public void damage(float amount, int killerId) {
         // Invoke entity damage event.
         EntityDamageEvent event = new EntityDamageEvent(this, amount, this.getScene().getEntityById(killerId));
         event.call();
-        if (event.isCanceled()) {
-            return; // If the event is canceled, do not damage the entity.
+        if (event.isCanceled() || !hasFightProperty(FightProperty.FIGHT_PROP_CUR_HP)) {
```
Birdulon: Check should be bundled in with the previous `this.getFightProperties() == null` rather than making an `EntityDamageEvent`.

src/main/java/emu/grasscutter/data/binout/ConfigGadgetCombatProperty.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.experimental.FieldDefaults;
+
+import java.io.IOException;
+
+@Data
+@FieldDefaults(level = AccessLevel.PRIVATE)
+public class ConfigGadgetCombatProperty {
+    @SerializedName(value="HP", alternate={"CMBBAPKBDCI"})
```
Birdulon: I suspect `HP` and `CMBBAPKBDCI` are different fields and `HP` having float and boolean values is a problem with with the resources rip.
image

A quick glance at my resources shows that no file contains both floats and bools for `HP`, and while I haven't checked all of the filenames against dim's repo, it appears that the float `HP` files are from 2.6.0 and the bool `HP` files are from 2.7.0, so I'd call it a wrong name translation on the 2.7.0 binoutput.

I think we'd be better off fixing resources than trying to work around this with gson adapters, especially since `"HP": true` also occurs on gadgets which aren't invulnerable.
image

Birdulon (changes requested):

src/main/java/emu/grasscutter/data/binout/ConfigGadgetCombatProperty.java
lilmayofuksu: dim...

Hartie95: Noticed later too, that there seems to be something wrong with the resources themself, after seeing some elements with both CMBBAPKBDCI  and HP.
I updated the branch to remove the TypeAdapter but kept CMBBAPKBDCI as alternate for HP for now.

src/main/java/emu/grasscutter/game/entity/GameEntity.java
Hartie95: fixed

Hartie95:
Updated the pull requests. It now also adds a lockHP Property to GameEntity and allow destroying of entities if its enabled, but the damage dealt is higher then the entites hp.

To full work, it now also needs these changes to the resources:
https://github.com/tamilpp25/Grasscutter_Resources/pull/15

Birdulon (approved):

tamilpp25:
> Updated the pull requests. It now also adds a lockHP Property to GameEntity and allow destroying of entities if its enabled, but the damage dealt is higher then the entites hp.
> 
> To full work, it now also needs these changes to the resources: [tamilpp25/Grasscutter_Resources#15](https://github.com/tamilpp25/Grasscutter_Resources/pull/15)

pushed to resources repo 👍🏻

PR #1777, opened 2022-09-14
Extend spawn command

Hartie95:
## Description
This extends the spawn command to allow better testing and debuging, while it also lets it use the same pattern as give and clear.
It also fixes the type of non vehicle entities spawned via spawn Command, adds some missing EntityTypes and moves some common patterns and methods from some commands to a new CommandHelpers class.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/ClearCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/ClearCommand.java
@@ -6,13 +6,15 @@ import emu.grasscutter.game.inventory.GameItem;
 import emu.grasscutter.game.inventory.Inventory;
 import emu.grasscutter.game.inventory.ItemType;
 import emu.grasscutter.game.player.Player;
+import lombok.Setter;
 
 import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
-import static emu.grasscutter.command.CommandHelpers.lvlRegex;
-import static emu.grasscutter.command.CommandHelpers.matchIntOrNeg;
+import static emu.grasscutter.command.CommandHelpers.*;
 
 @Command(
     label = "clear",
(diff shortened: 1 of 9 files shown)


src/main/java/emu/grasscutter/command/commands/ClearCommand.java
```diff
@@ -8,28 +8,21 @@
...
+
 @Command(
     label = "clear",
     usage = {"(all|wp|art|mat) [lv<max level>] [r<max refinement>] [<max rarity>*]"},
     permission = "player.clearinv",
     permissionTargeted = "player.clearinv.others")
 public final class ClearCommand implements CommandHandler {
-    private static Pattern lvlRegex = Pattern.compile("l(?:vl?)?(\\d+)");  // Java doesn't have raw string literals :(
     private static Pattern refineRegex = Pattern.compile("r(\\d+)");
     private static Pattern rankRegex = Pattern.compile("(\\d+)\\*");
```
Birdulon:
```suggestion
```

May as well move the rest of the regexes into `CommandHelpers` too

Hartie95: sound good, I will do that later

GanyusLeftHorn:
Suggestion: Maybe we could also add the ability to spawn monsters (1) with a set amount of HP, which might be very large/infinite, (2) with the AI turned off. This would effectively make it possible to spawn a test dummy monster. 

Something like that would have been very useful while working on energy, for example, and would also be useful for testing if certain passives etc. work.

Hartie95:
> Suggestion: Maybe we could also add the ability to spawn monsters (1) with a set amount of HP, which might be very large/infinite, (2) with the AI turned off. This would effectively make it possible to spawn a test dummy monster.
> 
> Something like that would have been very useful while working on energy, for example, and would also be useful for testing if certain passives etc. work.

Sound certainly usefull, I will look into it.

Hartie95:
Updated the pr, it now allows settings the stats of monsters and gadgets, overwriting the ai Id of monsters(e.g. 12001001 would be a non attacking but moving monster) and moves the int parameter handling to the CommandHelpers.

src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
```diff
@@ -112,4 +186,34 @@ private Position GetRandomPositionInCircle(Position origin, double radius) {
         target.addX((float) (r * Math.cos(angle))).addZ((float) (r * Math.sin(angle)));
         return target;
     }
+
+    private static class SpawnParameters {
+        @Setter
```
Birdulon: Stuff like this generally makes me prefer annotations on the same line (i.e. `@Setter public int id;`)

Hartie95: the extra lines are mostly from intellijs autoformat, but I can change them back to one liners if wished.

Birdulon (approved):

src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
Birdulon: yes please!

My general view on it would be that class annotations should almost always be on a separate line, property/field annotations should be on the same line unless there's a crazy amount of them (which I'd probably set at 4+, since there's a few `@Transient @Getter @Setter private Foo foo`s in the codebase already and they look reasonable imo)

Hartie95: done

PR #1784, opened 2022-09-15
Add functionality in MonsterData

ChisatoNishikigi73:
## Description
Add some functionality in MonsterData.java

stopAi(This only seems to change the state at the time of generation)
setKillDropId
setSkin
setDescribeId
setBaseHp
setBaseAttack
setBaseDefense
setElecSubHurt
setGrassSubHurt
setWaterSubHurt
setWindSubHurt
setIceSubHurt
setPhysicalSubHurt
setCampID
setWeaponId

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/data/excels/MonsterData.java
+++ b/src/main/java/emu/grasscutter/data/excels/MonsterData.java
@@ -1,7 +1,5 @@
 package emu.grasscutter.data.excels;
 
-import java.util.List;
-
 import emu.grasscutter.data.GameData;
 import emu.grasscutter.data.GameResource;
 import emu.grasscutter.data.ResourceType;
@@ -9,17 +7,19 @@ import emu.grasscutter.data.ResourceType.LoadPriority;
 import emu.grasscutter.data.common.PropGrowCurve;
 import emu.grasscutter.game.props.MonsterType;
 
+import java.util.List;
+
 @ResourceType(name = "MonsterExcelConfigData.json", loadPriority = LoadPriority.LOW)
 public class MonsterData extends GameResource {
-	private int id;
-	
-	private String monsterName;
+    private int id;
+
+    private String monsterName;
     private MonsterType type;
     private String serverScript;
     private List<Integer> affix;
     private String ai;
     private int[] equips;
-    private List<HpDrops> hpDrops;
+    private List<emu.grasscutter.data.excels.MonsterData.HpDrops> hpDrops;
     private int killDropId;
     private String excludeWeathers;
     private int featureTagGroupID;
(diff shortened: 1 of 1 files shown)


Hartie95:
You could probably use the @Getter and @Setter annotations for adding the setters and moving the getter code out there.

Hartie95:
Also might not be a good idea to directly edit the resources data objects, instead it might be better to use values from there as base for new properties on the game entities and edit them.

src/main/java/emu/grasscutter/data/excels/MonsterData.java
```diff
@@ -1,25 +1,25 @@
...
+    private int id;
+
+    private String monsterName;
     private MonsterType type;
     private String serverScript;
     private List<Integer> affix;
     private String ai;
     private int[] equips;
-    private List<HpDrops> hpDrops;
+    private List<emu.grasscutter.data.excels.MonsterData.HpDrops> hpDrops;
```
Birdulon, line 22: Why?

src/main/java/emu/grasscutter/data/excels/MonsterData.java
```diff
@@ -42,159 +42,222 @@ public class MonsterData extends GameResource {
...
-	        return this.HpPercent;
-	    }
-	}
+
+    @Override
+    public int getId() {
+        return this.id;
+    }
+
+    public String getMonsterName() {
```
Birdulon, line 55: Just use Lombok `@Getter` and `@Setter`?

Birdulon (changes requested):
For what purpose?

src/main/java/emu/grasscutter/data/excels/MonsterData.java
ChisatoNishikigi73: I will change it. It's my ide's problem.

GanyusLeftHorn (commented):
Excel data classes are meant to represent static resources that we read when the server starts, and never touch again. In other words, they are immutable. They have no setters by design, not by accident. (Aside: The only reason why their fields aren't declared `final`, or why we don't use `record`s for them, is that GSON seems to have a problem with that.)

As such, I would reject this PR.

ChisatoNishikigi73:
> For what purpose?

@ GanyusLeftHorn :
Suggestion: Maybe we could also add the ability to spawn monsters (1) with a set amount of HP, which might be very large/infinite, (2) with the AI turned off. This would effectively make it possible to spawn a test dummy monster.

Something like that would have been very useful while working on energy, for example, and would also be useful for testing if certain passives etc. work.

I just want to learn the language better, and make a contribution to Grasscutter, if I can help others, nothing else

GanyusLeftHorn:
> I just want to learn the language better, and make a contribution to Grasscutter, if I can help others, nothing else

@Kawaa-qwq That's great! Contributions are very welcome. Please don't be discouraged if you get negative feedback to one of your PRs. 

If you are unsure about what exactly to contribute, or whether your approach for a particular contribution is good, you can come discuss it with us in the #development channel in the Discord.

PR #1807, opened 2022-09-26
Wish the player a happy birthday :)

AZthemute:
## Description
Adds simple functionality for the player's birthday mail. Opening as draft to get people's opinion on something before finishing the PR (even though I doubt anyone will read this lmao)

**Currently, the implementation I have in mind is not 1:1 identical to the official server.** However, the official server's implementation doesn't make much sense in the context of a private server. The official server's implementation is to send a different mail for every year since the start of the game (i.e. the first birthday mail is only available before the day of the first anniversary, the second birthday mail is only available before the day of the second anniversary and after the day of the first anniversary, etc.). I just don't want to write a ton of code then have my PR rejected because the maintainers aren't happy with its implementation.

The implementation I have in mind is to send a different mail for every year since the player created their account. If this implementation is not okay with maintainers, let me know.

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


GanyusLeftHorn:
> The implementation I have in mind is to send a different mail for every year since the player created their account. If this implementation is not okay with maintainers, let me know.

That's totally okay. We don't need to try mimicking official _that much_. ;)

As for your current implementation draft, some notes:
- Things should not be hardcoded (I know your comment already says that, just adding it for completeness).
- The current logic of checking whether the player is online _at the time the job happens to run_ is problematic. I'm actually not sure if I have to be online on the day of my birthday to get the mail, but certainly not at a specific time. ;)
    * IF the player has to be online on the date to get the birthday mail, I'd suggest moving the logic to player daily reset handling (`Player::doDailyReset`).
    * IF they _don't_ have to be online, simply remove the check. That would of course require fixing the issue of players not receiving the mail notification when they got an offline mail.

AZthemute:
> * The current logic of checking whether the player is online _at the time the job happens to run_ is problematic. I'm actually not sure if I have to be online on the day of my birthday to get the mail, but certainly not at a specific time. ;)

If you don't log in during your birthday day, the mail is gone. You have to log in or else the game just doesn't care about you.


> * IF the player has to be online on the date to get the birthday mail, I'd suggest moving the logic to player daily reset handling (`Player::doDailyReset`).

Not sure how to do that currently, I'll look into it more when I actually have time to myself and when I don't forget lol

Either way, thanks for confirming!

Bannyfelt13:
why not do it as a plugin?

AZthemute:
> why not do it as a plugin?

I know less about plugin making than I do about coding (not much) and the official server does this, so I deemed it fit for the main Grasscutter.

AZthemute:
Official server does have a small difference with the birthday mails from other ones... that being that the language the mail is written in changes based on the language your game is in _after_ the mail is sent. It will always be the same as your game language, unlike other mails which only care about the language at the time of being sent. Not sure if this is a special type of mail that we simply don't have in GC yet, but worth noting anyway ~~mods don't close this PR pls I have more important things to do for the next couple of weeks than write code for anime game server~~

Birdulon:
>It will always be the same as your game language, unlike other mails which only care about the language at the time of being sent.

It seems to correspond to textmap hashes `2225380994` (2020-2021), `4120542850` (2021-2022), and `2660624634` (2022-2023?). The cakes also change each anniversary.
It could be a special type of mail, but it might just be an escape sequence to insert a textmap string.

KingRainbow44:

PR #1287, opened 2022-06-17
Element Switch for Main-Char

dianwangwin:
## Description

Please carefully read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md) before making any pull requests.
And, **Do not make a pull request to merge into stable unless it is a hotfix. Use the development branch instead.**
## Type of changes

Add a Command to switch elements

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] My pull request is unique and no other pull requests have been opened for these changes
- [x] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [x] I am responsible for any copyright issues with my code if it occurs in the future.

PR #1819, opened 2022-10-02
add SwitchElement command

TomyJan:
## Description

Switching the traveler's element. 
Although a [similar PR](https://github.com/Grasscutters/Grasscutter/pull/1287) has been turned off, I think this command is still necessary, because most of the actions in the game can be done with commands at the moment, and this should not be an exception. 
The code was modified from the source code of a plugin from an unknown source in the community, so I can't credit the original contributor. 

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


Bannyfelt13:
i guess maybe just use plugin

src/main/java/emu/grasscutter/command/commands/SwitchElementCommand.java
```diff
@@ -0,0 +1,202 @@
...
+        return (avatarId == 10000005) ? this.boyId : this.girlId;
+    }
+
+    private static /* synthetic */ Element[] $values() {
+        return new Element[] { Element.elementless, Element.pyro, Element.hydro, Element.anemo, Element.cryo, Element.geo, Element.electro, Element.dendro };
+    }
+
+
+    }
+    private static Method getPositionMethod() {
```
tamilpp25, line 67: why version support? this isn't a plugin now right?

tamilpp25 (changes requested):

Penelopeep:
Yes, I see it's being marked as closed but let me say two things:
1. I'm the author of this plugin, and this one is a very old version.
2. I'm still up to adding it as Grasscutter command but this one and the old pr was closed "in favor of this plugin" so if you want to add this properly then let me know, I'll help you with my spaghetti code

PR #1826, opened 2022-10-04
Fix showing ban message box

phuchptty:
## Description
I have tested and found that, if we close the session right before sending the response packet, the client will show a "Disconnected" message instead of a ban. So I removed it.
After the client press OK or a time, the client would automatically close the session and disconnect.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/server/game/GameServerPacketHandler.java
+++ b/src/main/java/emu/grasscutter/server/game/GameServerPacketHandler.java
@@ -4,6 +4,7 @@ import static emu.grasscutter.config.Configuration.*;
 
 import java.util.Set;
 
+import emu.grasscutter.game.Account;
 import emu.grasscutter.server.event.game.ReceivePacketEvent;
 import org.reflections.Reflections;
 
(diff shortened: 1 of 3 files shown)


Birdulon:
The problem with leaving a banned session open is that all the handlers will still be running. A client that ignores the ban message (e.g. by running the kind of hacks that might earn them a ban in the first place) will be able to do continue playing normally.

phuchptty:
> The problem with leaving a banned session open is that all the handlers will still be running. A client that ignores the ban message (e.g. by running the kind of hacks that might earn them a ban in the first place) will be able to do continue playing normally.

I think I can add a delay to close the session. Delay about 2 seconds then close

GanyusLeftHorn:
I think the cleaner solution would be to introduce a new `SessionState` that disables handling of packets after being set. See `GameServerPacketHandler::handle()` for reference.

phuchptty:
> I think the cleaner solution would be to introduce a new `SessionState` that disables handling of packets after being set. See `GameServerPacketHandler::handle()` for reference.

Yeah. That was [Birdulon](https://github.com/Birdulon) tell me in discord too

Birdulon (approved):

PR #1836, opened 2022-10-07
Add sqlite support

WhichWho:
## Description

Add sqlite support.
Refactor BaseDatabase interface for more db support extension.

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


WhichWho:
edite config.json databaseInfo connectionUri 
```
jdbc:sqlite:grasscutter.db
```
switch to sqlite database

WhichWho:
> In principle, I like the idea of generalizing our database layer to allow for different types of databases. However, in your current implementation, you support SQLite by generating plain SQL statements directly in your code. That is bad practice and shouldn't be done anymore.
> 
> I'd suggest looking into proper ORM for SQLite (or, better yet, use one that already supports multiple database vendors out of the box), and changing the PR to use that instead.

Maybe Mybatis is a batter choice. I will try it later.

Birdulon:
`Mybatis` looks like a better choice, pure java and small lib size.

`sqlite-jdbc` has JNI builds included and the docs weren't clear on if the pure Java sqlite fallback even exists on current versions, seemed to suggest it was only on a ~2009 version. It would also add 12MB to the GC jar.

WhichWho:
I tried a lot of ways. Unfortunately. the raw use of sqlite driver is necessary. 
After use Mybatis(or other ORM else). I can't find better ways to  deserialization the complex type.(nesting java beans, List, Set, Map with generic type)
Maybe two ways can handle the problem.

1. save the collections inner generic type with data together.

2. rewrite all the @Entity class, flat the data formate to support SQL.

But neither of ways is good choice.
Current sqlite implemention used lots of TypeToken to force the complex type deserialization correct.

Finally, maybe mongodb can not be replaced. It had been high coupled with whole project.

KingRainbow44:
As mentioned, **Morphia** is *heavily* integrated with Grasscutter. Adding SQLite support at this time is something that would require a lot of breaking changes across the whole server. For this reason, I will be closing this PR.

PR #1839, opened 2022-10-07
Move Data, Plugin, Script, Packet access from Strings to Paths

Birdulon:
## Description
- No longer dump default Data files to folder on launch
- Allow Scripts to be loaded from Resources zip
- Lay groundwork for Plugins to be loaded from zip

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/config/ConfigContainer.java
+++ b/src/main/java/emu/grasscutter/config/ConfigContainer.java
@@ -84,7 +84,7 @@ public class ConfigContainer {
         public String resources = "./resources/";
         public String data = "./data/";
         public String packets = "./packets/";
-        public String scripts = "./resources/Scripts/";
+        public String scripts = "resources:Scripts/";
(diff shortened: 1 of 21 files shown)


eternalcomet:
I think it may be necessary to allow developer to dump default Data files to folder,because it is more convenient to check or modify some config files.We may implement this by adding  a switch in config.json or adding a command to do that.

eternalcomet:
We have two solutions for loading resources in IDE running:
1.use `Grasscutter.class.getResource()` instead of `Grasscutter.class.getProtectionDomain().getCodeSource().getLocation().toURI()`;
(This is also what grasscutter did before)
2.dump all data files if finding `Grasscutter.class.getProtectionDomain().getCodeSource().getLocation().toURI()` is a path rather than a jar file.

PR #1840, opened 2022-10-07
Update SpawnCommand.java

Hiro420:
## Description

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
@@ -25,7 +25,6 @@ import static emu.grasscutter.utils.Language.translate;
 
 @Command(
     label = "spawn",
-    aliases = {"spawn"},
     usage = {
         "<itemId> [x<amount>] [blk<blockId>] [grp<groupId>] [cfg<configId>] <x> <y> <z>",
         "<gadgetId> [x<amount>] [state<state>] [maxhp<maxhp>] [hp<hp>(0 for infinite)] [atk<atk>] [def<def>] [blk<blockId>] [grp<groupId>] [cfg<configId>] <x> <y> <z>",
--- a/src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
@@ -25,7 +25,7 @@ import static emu.grasscutter.utils.Language.translate;
 
 @Command(
     label = "spawn",
-    aliases = {"drop"},
+    aliases = {"spawn"},
     usage = {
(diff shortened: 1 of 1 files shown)


Hartie95:
Why do you want to remove the drop alias and why won't you remove the alias line completely and instead set the command itself as alias?

Hiro420:
> Why do you want to remove the drop alias and why won't you remove the alias line completely and instead set the command itself as alias?

well its spawn command not drop command. i saw a lot of people struggling to spawn enemies because instead of /spawn <enemyid> it became something like /drop <enemyid> which is kinda stupid for a spawn command

Hiro420:
also its a very simple and quick fix. let's say that the problem was fixed, wasn't it?

Hartie95:
> > Why do you want to remove the drop alias and why won't you remove the alias line completely and instead set the command itself as alias?
> 
> well its spawn command not drop command. i saw a lot of people struggling to spawn enemies because instead of /spawn it became something like /drop which is kinda stupid for a spawn command

But both (/spawn and /drop) work, and it can also be used to create drops and gadgets, in which case drop makes sense.
Also if you would remove the alias, it would one, not be bugfix, and two, you would remove the complete alias line, not replace it with what the command already is, since the command itself is not an alias.

Hartie95:
Instead of removing drop, you could add a shorter alias, that makes sence and is less confusing, since the help description always uses the shortest alias.
Maybe something like `spn` or `sp` as alias would be less confusing.

Hiro420:
ok now its enough? XD

Hartie95:
> ok now its enough? XD

Besides the info about either removing it fully, or adding another one, it was all suggestions from my side. And in the end the grasscuter team decides what gets merged, not me. 
Also if you add an alias thats shorter then drop, there is no need to also remove drop, you could have both aliases. ^^

Birdulon:
> i saw a lot of people struggling to spawn enemies because instead of /spawn it became something like /drop which is kinda stupid for a spawn command

It's not though?
image
image

As Hartie pointed out, it's still very much "drop" in the context of items. It could probably have `s` as an alias but I see no reason to remove the `drop` alias.

src/main/java/emu/grasscutter/command/commands/SpawnCommand.java
```diff
@@ -25,7 +25,7 @@
 
 @Command(
     label = "spawn",
-    aliases = {"drop"},
+    aliases = {"spn"},
```
Birdulon, line 28:
```suggestion
    aliases = {"s", "drop"},
```

Birdulon (changes requested):

PR #1841, opened 2022-10-08
fix the JVM memory issue for people with low ram

Hiro420:
is to fix the JVM memory issue for some people with low amount of ram

## Description

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,1 @@
+org.gradle.jvmargs=-Xmx4096M


Birdulon:
I'm not sure what issue this is supposed to fix, the default value for that is 512MB so raising it to 4096MB is likely to be worse for low RAM.
https://docs.gradle.org/5.6.4/userguide/build_environment.html#sec:configuring_jvm_memory

KingRainbow44:
More specifically, this would only help people when *building* Grasscutter. It's pretty much assumed that if you're building Grasscutter that you either: A: have enough ram or B: know how to allocate more JVM heap in your system's **global** `gradle.properties` file.

TLDR: Instead of proposing this *directly* to the repo, it's a better idea to instead add this to documentation.

PR #1842, opened 2022-10-08
fix costumes are unable to be unlocked.

hamusuke0323:
## Description

## Type of changes

- [x ] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

- [ x] My code follows the style guidelines of this project
- [ x] I have read the [Contributing note](https://github.com/Grasscutters/Grasscutter/blob/stable/CONTRIBUTING.md) and [Code of conduct](https://github.com/Grasscutters/Grasscutter/blob/development/CODE_OF_CONDUCT.md)
- [ x] I am responsible for any copyright issues with my code if it occurs in the future.

### DIFF
--- a/src/main/java/emu/grasscutter/data/excels/AvatarCostumeData.java
+++ b/src/main/java/emu/grasscutter/data/excels/AvatarCostumeData.java
@@ -1,11 +1,13 @@
 package emu.grasscutter.data.excels;
 
+import com.google.gson.annotations.SerializedName;
 import emu.grasscutter.data.GameData;
 import emu.grasscutter.data.GameResource;
 import emu.grasscutter.data.ResourceType;
 
 @ResourceType(name = "AvatarCostumeExcelConfigData.json")
 public class AvatarCostumeData extends GameResource {
+    @SerializedName(value = "skinId", alternate = "costumeId")
     private int skinId;
     private int itemId;
     private int characterId;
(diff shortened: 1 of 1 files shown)


hamusuke0323:
We cant get avatar costumes because the property name "costumeId" was changed to "skinId" in AvatarCostumeExcelConfigData.json from V3.1 update probably, I guess.

src/main/java/emu/grasscutter/data/excels/AvatarCostumeData.java
```diff
@@ -6,14 +6,14 @@
 
 @ResourceType(name = "AvatarCostumeExcelConfigData.json")
 public class AvatarCostumeData extends GameResource {
-    private int costumeId;
+    private int skinId;
```
Birdulon, line 11:
```suggestion
    @SerializedName(value = "skinId", alternate = "costumeId")
    private int skinId;
```

Birdulon:
Looks like `skinId` is the value from 2.2 NT, we should keep `costumeId` as an alternate name so that old resources don't break for now. (probably remove the alternate name in a patch or two)

hamusuke0323:
I have a unrelated question. May I contribute to gc with using generated .java files from Sorapointa .proto files?

Birdulon:
Yes, that's acceptable

hamusuke0323:
I see. thanks for your reply

hamusuke0323:
sorry, I forgot splitting PR. should I revert?

Birdulon:
Make a new branch at your current `head` (`git branch namecards`) and reset your `development` branch to before those commits (`git reset --hard 9182835`)

Birdulon:
note that you'll also have to force push to `development` (`git push -f`)

Birdulon (approved):

PR #1845, opened 2022-10-10
fix: albedo elevator doesn't work

hamusuke0323:
## Description

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


hamusuke0323:
image
it seems to works, but the player won't rise. probably need to call some scripts.

phuchptty:
real elevator

akatatsu27:
> image it seems to works, but the player won't rise. probably need to call some scripts.

at least it will never gonna let you down

src/main/java/emu/grasscutter/server/packet/send/PacketPlatformStartRouteNotify.java
```diff
@@ -0,0 +1,39 @@
...
+                .setPosOffset(entity.getPosition().toProto())
+                .setRotOffset(MathQuaternionOuterClass.MathQuaternion.newBuilder()
+                    .setW(1.0F)
+                    .build())
+                .setMovingPlatformType(entity.getMovingPlatformType())
+                .setIsActive(true)
+                .build())
+            .build();
+
+        entity.setStarted(true);
```
Hartie95: Wouldn't it be better to have something like entity.startRoute() that sets these things and sends this package, instead of letting the notifypackage set entity proppeties ?

akatatsu27: Agreed.  send Packets only have constructor(s) because they are used just as data for the server to send. Since entities are scene-bound objects, platforms should be managed by the scene they are in. This is especially important for seelie (they are considered platforms) which have complex routes and behaviour

Hartie95:
> it seems to works, but the player won't rise. probably need to call some scripts.

Its doesn't seem to be a script problem, instead the server just doesn't update the scene time, if scene.getTime is even intended to represent the sceneTime.
I got your branch to work by adding these to the scene.java and use scene.getSceneTime instead if scene.getTime:
```java
    private long startTime;

    public Scene(World world, SceneData sceneData) {
    ...
        this.startTime = System.currentTimeMillis();
    ...
    }
    
    public int getSceneTime(){
        return (int)( currentTimeMillis()-startTime);
    }
```

Also it might maybe be good to let the GameEntity generate the PlatformInfo, since its used in multiple packages, and always based on informations of the entity/gadget

src/main/java/emu/grasscutter/server/packet/send/PacketPlatformStartRouteNotify.java
hamusuke0323: ok, I will fix it!

hamusuke0323:
> > it seems to works, but the player won't rise. probably need to call some scripts.
> 
> Its doesn't seem to be a script problem, instead the server just doesn't update the scene time, if scene.getTime is even intended to represent the sceneTime. I got your branch to work by adding these to the scene.java and use scene.getSceneTime instead if scene.getTime:
> 
> ```java
>     private long startTime;
> 
>     public Scene(World world, SceneData sceneData) {
>     ...
>         this.startTime = System.currentTimeMillis();
>     ...
>     }
>     
>     public int getSceneTime(){
>         return (int)( currentTimeMillis()-startTime);
>     }
> ```
> 
> Also it might maybe be good to let the GameEntity generate the PlatformInfo, since its used in multiple packages, and always based on informations of the entity/gadget

No wonder the elevator was appeared above the player suddenly not raise the player. Thanks for your suggestion.

hamusuke0323:
image
probably it works.

phuchptty:
is it work the same with gadget elevator or automatic door ?

hamusuke0323:
> is it work the same with gadget elevator or automatic door ?

what is automatic door? at least, albedo elevator will raise the player.

alt3ri:
GIGACHAD HOLY

nitrog0d:
https://user-images.githubusercontent.com/31241699/195163105-2bde2a00-86f9-46b7-8ac7-56cdef48b9ec.mov

Birdulon (changes requested):
Looking good!

In general we try to avoid referencing character names in the code. Renaming `EntityAlbedoSolarIsotomaClientGadget` to `EntitySolarIsotomaClientGadget`, and `EntityAlbedoElevatorPlatform` to `EntitySolarIsotomaElevatorPlatform` should be good.

Birdulon (approved):

PR #1864, opened 2022-10-16
feat: set emoji collection in chat screen

hamusuke0323:
## Description

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


src/main/java/emu/grasscutter/game/player/Player.java
```diff
@@ -1187,6 +1189,7 @@ public void onLogin() {
         session.send(new PacketAllWidgetDataNotify(this));
         session.send(new PacketWidgetGadgetAllDataNotify());
         session.send(new PacketCombineDataNotify(this.unlockedCombines));
+        session.send(new PacketGetChatEmojiCollectionRsp(this.getChatEmojiIdList()));
```
akatatsu27, line 1192: Rsp packets should only be sent as a response to the corresponding Req packets. You already send GetChatEmojiCollectionRsp via HandlerGetChatEmojiCollectionReq::handle, so this is redundant at best
```suggestion

```

akatatsu27 (changes requested):

Birdulon (approved):
Testing note: Client seems to cache emoji collection, even when relogging, which may give the impression that the feature is working unless you actually close and reopen the client. This PR makes it actually work.

src/main/java/emu/grasscutter/game/player/Player.java
hamusuke0323: you are right, but client doesn't seem to send request packet.

akatatsu27: Then the issue is elsewhere. Perhaps the client expects some prior notification packet from the server for it to then send the Req packet (as is the case with ActivityScheduleInfoNotify and GetActivityInfoReq). Rsp packets with no prior Req shouldn't happen.

hamusuke0323: I will try to find the prior packet.

hamusuke0323: I noticed that `PacketPlayerLoginRsp` wasn't sent. Because of this, the request didn't come probably.

PR #1865, opened 2022-10-16
Extend give command "skill level" & shortening /talent all

natsurepo:
## Description

Attempting to add talents / skill level as a give command parameter, also shortening /talent all command
- [x] ~~Adding getCombatSkills for grabbing skill ids specifically for talent command, since getSkillsAndEnergySkill also grabs other unrelated skills~~ (removed since getSkillsAndEnergySkill is enough and there is a level cap for invalid talents)
- [x] Shorten /talent all code
- [x] Adding skillLevel as parameter so people can also specify talent levels for characters in /give
- [x] Resolving the issue for skillLevel parameter messing with constellations not giving extra skillLevel at the end (ex. at C6, talent E and Q doesn't increase by 3) (resolved by [1ecc3f4](https://github.com/Grasscutters/Grasscutter/commit/1ecc3f43905d4d29f409d86654aa113bc948b2af))
- [x] Cleanup 

A bit of note in the end, I would appreciate any suggestions and help for either resolving my current issue and making the implementation tidier.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


src/main/java/emu/grasscutter/game/avatar/Avatar.java
```diff
@@ -779,6 +779,15 @@ public boolean setSkillLevel(int skillId, int level) {
         return true;
     }
 
+    public void changeSkillLevel(int skillLevel) {
+        if (skillLevel < 1) skillLevel = 1;
+        if (skillLevel > 12) skillLevel = 12; // Set max to 12 at the moment instead of 15 (it will conflict with constellation's talent level bonus)
```
Birdulon: Check `Avatar::setSkillLevel`, constellation bonuses have also been capped in 4e75361 which also correctly caps other talents (e.g. alt sprints) so it would be safe (even if slightly incorrect) to just use `AvatarSkillDepotData::getSkillsAndEnergySkill` elsewhere.

src/main/java/emu/grasscutter/data/excels/AvatarSkillDepotData.java
```diff
@@ -77,4 +77,10 @@ public IntStream getSkillsAndEnergySkill() {
         return IntStream.concat(this.skills.stream().mapToInt(i -> i), IntStream.of(this.energySkill))
                         .filter(skillId -> skillId > 0);
     }
+
+    public IntStream getCombatSkills() {
+        return IntStream.concat(IntStream.of(this.getSkills().get(0), this.getSkills().get(1)), IntStream.of(this.energySkill))
```
Birdulon:
```suggestion
        return IntStream.of(this.getSkills().get(0), this.getSkills().get(1), this.energySkill)
```

src/main/java/emu/grasscutter/game/avatar/Avatar.java
natsurepo: I added `Avatar::changeSkillLevel` specifically for my implementation since I only needed skillLevel as input (instead of also needing skillId), also using `Avatar::setSkillLevel` leads to `"emu.grasscutter.game.avatar.Avatar.getPlayer()" is null`.

Also for the `AvatarSkillDepotData::getSkillsAndEnergySkill` does it mean I should revert my changes for `AvatarSkillDepotData` and `TalentCommand`?

Birdulon: >also using `Avatar::setSkillLevel` leads to `"emu.grasscutter.game.avatar.Avatar.getPlayer()" is null`.

Ah right, won't have a player during construction. Should be fine to add a conditional block to the packets in it.
```java
        // Packet
        val player = this.getPlayer();
        if (player != null) {
            player.sendPacket(new PacketAvatarSkillChangeNotify(this, skillId, oldLevel, level));
            player.sendPacket(new PacketAvatarSkillUpgradeRsp(this, skillId, oldLevel, level));
        }
```

> Also for the `AvatarSkillDepotData::getSkillsAndEnergySkill` does it mean I should revert my changes for `AvatarSkillDepotData` and `TalentCommand`?

Up to you, might be useful regardless and it's a pretty harmless addition.

> I added `Avatar::changeSkillLevel` specifically for my implementation since I only needed skillLevel as input 

This is a bit overspecialised though.

natsurepo:
```java
public boolean setSkillLevel(int skillId, int level) {
    if (level < 0 || level > 15) return false;
    var validLevels = GameData.getAvatarSkillLevels(skillId);
    if (validLevels != null && !validLevels.contains(level)) return false;
    int oldLevel = this.skillLevelMap.getOrDefault(skillId, 0);
    this.skillLevelMap.put(skillId, level);
    this.save();
    // Packet
    val player = this.getPlayer();
    if (player != null) {
        player.sendPacket(new PacketAvatarSkillChangeNotify(this, skillId, oldLevel, level));
        player.sendPacket(new PacketAvatarSkillUpgradeRsp(this, skillId, oldLevel, level));
    }
    return true;
}

public void changeSkillLevel(int skillLevel) {
    int finalSkillLevel = skillLevel;
    skillDepot.getCombatSkills().forEach(id -> this.setSkillLevel(id, skillLevel));
}
```

Made some changes with your suggestion and it works with `setSkillLevel` now. I still keep `changeSkillLevel` as an intermediate operation to fetch skillIds. Still no luck with bonus talent levels from the constellations though.

Birdulon:
Added cbd66a74 after noticing the `sl(\d+)` regex was getting beaten out by the existing `l(\d+)` match for level

src/main/java/emu/grasscutter/command/commands/GiveCommand.java
```diff
@@ -212,30 +214,29 @@ public void execute(Player sender, Player targetPlayer, List<String> args) {
...
 
-    private static Avatar makeAvatar(AvatarData avatarData, int level, int promoteLevel, int constellation) {
+    private static Avatar makeAvatar(AvatarData avatarData, int level, int promoteLevel, int constellation, int skillLevel) {
         Avatar avatar = new Avatar(avatarData);
         avatar.setLevel(level);
         avatar.setPromoteLevel(promoteLevel);
+        avatar.changeSkillLevel(skillLevel);
         avatar.forceConstellationLevel(constellation);
-        avatar.recalcStats();
+        avatar.recalcConstellations();
```
Birdulon:
```suggestion
```
`Avatar::forceConstellationLevel` takes care of this. It also handles `recalcStats`, but it might not call that if it's fed an invalid constellation level so leaving `recalcStats` in is fine.

natsurepo: Alright, thank you

src/main/java/emu/grasscutter/game/avatar/Avatar.java
```diff
@@ -762,11 +763,19 @@ public boolean setSkillLevel(int skillId, int level) {
...
-        this.getPlayer().sendPacket(new PacketAvatarSkillUpgradeRsp(this, skillId, oldLevel, level));
+        val player = this.getPlayer();
+        if (player != null) {
+            player.sendPacket(new PacketAvatarSkillChangeNotify(this, skillId, oldLevel, level));
+            player.sendPacket(new PacketAvatarSkillUpgradeRsp(this, skillId, oldLevel, level));
+        }
         return true;
     }
 
+    public void changeSkillLevel(int skillLevel) {
```
Birdulon: Still not too keen on this method being added.

src/main/java/emu/grasscutter/game/avatar/Avatar.java
```diff
@@ -753,7 +753,8 @@ public boolean upgradeSkill(int skillId) {
     }
 
     public boolean setSkillLevel(int skillId, int level) {
-        if (level < 0 || level > 15) return false;
+        if (level < 1) level = 1;
+        if (level >15) level = 15;
```
Birdulon: Shouldn't cap levels here, also setting to 0 is a feature that shouldn't be removed.

natsurepo:
```java
    private static void giveAllAvatars(Player player, GiveItemParameters param) {
        int promoteLevel = Avatar.getMinPromoteLevel(param.lvl);
        if (param.constellation < 0 || param.constellation > 6) param.constellation = 6; // constellation's default is -1 so if no parameters set for constellations it'll automatically be 6
        if (param.skillLevel > 15) param.skillLevel = 15;
        if (param.skillLevel < 0) param.skillLevel = 1; // Still allowing 0 to be used, while everything below 0 goes back to default
        for (AvatarData avatarData : GameData.getAvatarDataMap().values()) {
            int id = avatarData.getId();
            if (id < 10000002 || id >= 11000000) continue; // Exclude test avatars
            // Don't try to add each avatar to the current team
            player.addAvatar(makeAvatar(avatarData, param.lvl, promoteLevel, param.constellation, param.skillLevel), false);
        }
    }
```
Can I move it here instead or remove it completely?

Birdulon: Just revert the two line change :)

src/main/java/emu/grasscutter/game/avatar/Avatar.java
natsurepo: Removed `changeSkillLevel`, added -1 as input for `skillId` in `GiveCommand` and `setSkillLevel` will fetch those ids within itself if `skillId` is -1

natsurepo: I tried it in game, no conflicts with other usages too

Birdulon (approved):

PR #1884, opened 2022-10-22
Extend setConstCommand "all"

natsurepo:
## Description

This PR attempts to extend the usage of `setConstCommand`, now players can target all existing avatars in their inventory.
- [x] Added parameters "all" to target all currently owned avatar
- [x] Create `reloadScene` to remove duplicate code (it's used by `setConstellation` and `setAllConstellation`)
- [x] Added `setAllConstellation` as a separate method since `setConstellation` for all avatars would trigger `reloadScene` multiple times in some cases
- [x] Added new translated message `commands.setConst.successall`
- [x] Cleanup (remove unused imports)

P.S. I think I have github skill issue since it's 18 commits 💀 

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


TukanDev:
wonderhoy

natsurepo:
wonderhoy these 17 unrelated commits 😭

src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
```diff
@@ -45,7 +63,7 @@ public void execute(Player sender, Player targetPlayer, List<String> args) {
 
             CommandHandler.sendTranslatedMessage(sender, "commands.setConst.success", avatar.getAvatarData().getName(), constLevel);
         } catch (NumberFormatException ignored) {
-            CommandHandler.sendTranslatedMessage(sender, "commands.setConst.level_error");
+            CommandHandler.sendTranslatedMessage(sender, "commands.setConst.fail");
```
Birdulon: ?

src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
```diff
@@ -30,6 +29,25 @@ public void execute(Player sender, Player targetPlayer, List<String> args) {
             return;
         }
 
+        if (args.get(0).equalsIgnoreCase("all")) {
+            if (args.size() < 2) {
+                sendUsageMessage(sender);
+                return;
+            }
+            try {
+                int constLevel = Integer.parseInt(args.get(1));
```
Birdulon: Consider using `args.remove(index)` to avoid duplicating the level parsing.
Alternatively, move the "all" to the second argument and take inspiration from the `switch` handling in other commands.

Birdulon (approved):

src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
natsurepo: `level_error` is only relevant before I added `all`,

```java
if (args.get(0).equalsIgnoreCase("all"))
```

This `if` few lines before is used to check if the player inputs `all`, but if they use any other word it won't make sense getting a response of `Constellation level must be between 0 and 6`, so I switched it to a generic fail message

src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
natsurepo: > Alternatively, move the "all" to the second argument and take inspiration from the switch handling in other commands.

I agree on that one

natsurepo:
Should be all set

src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
```diff
@@ -1,25 +1,24 @@
...
 import java.util.List;
 
 @Command(
     label = "setConst",
     aliases = {"setconstellation"},
-    usage = {"<constellation level>"},
+    usage = {
+        "<constellation level>",
+        "<constellation level> all"
+    },
```
Birdulon:
```suggestion
    usage = {"<constellation level> [all]"},
```

Birdulon (changes requested):

Birdulon (approved):
Avatars outside the party don't update clientside until they are added to an active team, I don't see a feasible way to handle that for now, though.

PR #1892, opened 2022-10-24
Implement alias system

natsurepo:
## Description

This PR attempts to implement alias system for players to set for themselves. Inspired by source dev console `alias` feature.

- [x] Add `AliasCommand.java` as an interface for players (arguments checks and majority of the feedback messages were handled here)
- [x] Add `AliasManager.java` to handle database changes
- [x] Add a new map in `Player.java` to store set aliases
- [x] Add alias handling so it is registered when player do input them
- [x] Resolve conflicts of overlapping alias and command (when they have the same name)
- [ ] Add global aliases
- [ ] Any sub features would be added here
- [ ] Swap messages with translated messages and language stuffs (finally no lint failure)
- [ ] Cleanup

P.S. Please do let me know if there are any suggestions or ideas that might result in better (and tidier) implementation of this feature

## Usages
`/alias set <alias> [commands]` Store the alias and commands inside player's database
`/alias info <alias>` Fetches the alias and commands for said alias
`/alias clear <alias>` Clears the said alias from player's database
`/alias list` Lists all the set aliases for the target player
`/alias clearall` Clears all the set aliases for the target player

Example:
`/alias set natsu prop um 1; prop ue 1; prop us 1; give avatars lv90 c6 sl10; give 102 1000000`
Makes a new map entry in `CommandAliases` with `natsu` as key, and list of
`[prop um 1, prop ue 1, prop us 1, give avatars lv90 c6 sl10, give 102 1000000]` inside the player's database

Inputting `/natsu` from said player will execute these commands:
- /prop um 1
- /prop ue 1
- /prop us 1
- /give avatars lv90 c6 sl10
- /give 102 1000000

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


Birdulon:
>Add alias handling so it is registered when player do input them

This will need changes in `CommandMap` regardless, it would be better to localise the code to there, just as the target handling code currently is. A new Manager class for this is complete overkill.

GanyusLeftHorn:
Looking at the example you provided, this seems to not only be a system for aliases but also for batching commands. That's a pretty useful thing to have, and probably not just on the player level. I could see server operators wanting to use this to provide useful batch commands for all their players. Maybe this feature could be extended to also allow defining global aliases that are visible to all players?

natsurepo:
> Maybe this feature could be extended to also allow defining global aliases that are visible to all players?

Yes it is listed within my to do list, but the details of it (how to do that in the best way) are still up for discussion

KingRainbow44:

PR #1913, opened 2022-10-29
Fix Mail SendToAll Freeze

jie65535:
## Description

> Sending an mail to everyone in a 4h8g server with 3316 accounts would cause the server to use up all resources and then crash.

Calling the `DatabaseHelper.getAllPlayers()` method is too expensive, changing to a `Stream` has a significant improvement.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/database/DatabaseHelper.java
+++ b/src/main/java/emu/grasscutter/database/DatabaseHelper.java
@@ -1,6 +1,7 @@
 package emu.grasscutter.database;
 
 import java.util.List;
+import java.util.stream.Stream;
 
 import com.mongodb.client.result.DeleteResult;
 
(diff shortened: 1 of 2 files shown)


jie65535:
This improves the send method to avoid crashes, but it still takes a lot of time to perform the action, maybe you can start a thread to continue and let the method return first.

src/main/java/emu/grasscutter/database/DatabaseHelper.java
```diff
@@ -154,8 +155,8 @@ public static synchronized void deleteAccount(Account target) {
         DatabaseManager.getAccountDatastore().find(Account.class).filter(Filters.eq("id", target.getId())).delete();
     }
 
-    public static List<Player> getAllPlayers() {
-        return DatabaseManager.getGameDatastore().find(Player.class).stream().toList();
+    public static Stream<Player> getAllPlayers() {
```
Birdulon: Sadly, this is an API break. This will need to be a new method, and unfortunately it can't share the name `getAllPlayers`. Since the name conflict is a problem anyway, it might be better to just make something like:

```java
public static <T> Stream<T> getByGameClass(Class<T> classType) {
    return DatabaseManager.getGameDatastore().find(classType).stream();
}
```

jie65535: ok

jie65535: By the way, I recommend annotating the method as deprecated and explaining why.

Birdulon: Yeah, `@Deprecated(forRemoval = true)` will do it.

Birdulon (approved):

PR #1914, opened 2022-10-30
Move reloadScene() to the Player class to allow call it from another …

BiosNod:
## Description

Move reloadScene() to the Player class to allow call it from another places (not only from the SetConstCommand class)

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/SetConstCommand.java
@@ -58,7 +58,7 @@ public final class SetConstCommand implements CommandHandler {
 
         // force player to reload scene when necessary
         if (constLevel < currentConstLevel) {
-            this.reloadScene(player);
+            player.reloadScene();
         }
 
         // ensure that all changes are visible to the player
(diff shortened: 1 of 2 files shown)


GanyusLeftHorn:
Maybe this would make more sense in `Scene` rather than `Player`?

Birdulon:
> Maybe this would make more sense in `Scene` rather than `Player`?

The issue is that it doesn't actually reload the scene, it just hackily gets the player to accept update packets that are relevant to the active party for things that don't have normal incremental state updates like down-levelling talents and removing constellations. I'd prefer not to codify it anywhere, really.

Birdulon:
> I'd prefer not to codify it anywhere, really.

Just to follow up on that - we might find a better hack for talents and/or constellations refresh. If it's a public function, we can't change that to the new hack in case some plugin is using it for something else that only the old hack works for. We're then stuck with a function that doesn't really make sense to have, and removing it is an API break.

PR #1962, opened 2022-11-23
TSJ and TSV parsing

Birdulon:
## Description
Excel-style classes can now be loaded from two different Tab-separated table formats:

### TSJ - Tab-Separated JSON
A hybrid format where non-primitive fields are represented with JSON string values.
e.g.
image

### TSV - Tab-Separated Value
A more traditional format where the columns all represent primitive values. Complex objects will have flattened keys to accommodate this.
e.g.
image
_The deserializer for TSV is currently about 25x slower than the TSJ and JSON deserializers, as such it is not recommended for general use right now. Some of you may have some files in this format already, so this may be convenient to you if you don't mind waiting a few seconds for everything to load instead of half a second._

### Purpose
JSON is a relatively wasteful format for table-style data. When most of the objects all have the same fields, proper table formats can be substantially smaller uncompressed:
image
While the size difference when compressed is much smaller (a few MB rather than over a hundred), this still has advantages for maintaining resource repositories, and in most cases, produces cleaner diffs when no columns are added.
It's also convenient to view and edit the files in a Spreadsheet editor such as Excel. I intend to move some default `data` files over to TSJ format for this reason, starting with `banners.json` and then doing the gigantic spawns and drops files (which will be updated in the process).

`ExcelBinOutput` resources will load in the following priority: TSJ > JSON > TSV.
The current resources repo uses JSON for everything, and anyone interested in converting JSON files to optimized TSJ and TSV files can use my script https://github.com/Birdulon/GrasscutterTools/blob/master/excel_json_to_tsv.ipynb or make their own, it's a fairly easy conversion after all!

## Type of changes
- [ ] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/data/ResourceLoader.java
+++ b/src/main/java/emu/grasscutter/data/ResourceLoader.java
@@ -114,7 +114,7 @@ public class ResourceLoader {
         getResourceDefClassesPrioritySets().forEach(classes -> {
             // Grasscutter.getLogger().error("Loading resources: " + classes);
             classes.stream()
-                .parallel().unordered()
+                // .parallel().unordered()
                 .forEach(c -> {
                     val type = c.getAnnotation(ResourceType.class);
                     if (type == null) return;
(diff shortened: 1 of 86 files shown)


Birdulon:
Oh yeah, this also parallelizes `ExcelBinOutput` loading in general, so people should see a modest improvement in startup time.

PR #1987, opened 2022-12-20
Add a new command 'unlockmap' to unlock map for players

360NENZ:
## Description
## Issues fixed by this PR
Add a new command `unlockmap` to unlock map for players

**Why was this command added?**
If you are a server owner, your server imitates the official server, and you need to work yourself to obtain materials, and you want to control the permissions of the players, so that they can only unlock the map, and cannot use all commands except `unlockmap`, because `prop` command it can not only unlock the map, but also execute other commands, so there is this separate command.

You can also add the permission `player.unlockmap` in config.json!
Like this: 
```
"defaultPermissions": [player.unlockmap]
```
**Features:**

- Unlock map commands independent of props
- Usage: `unlockmap @<UID>`
- Alias: `um @<UID>`
- Permission: `player.unlockmap` (Permission `player.unlockmap.others` is also required to use on other players)

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


TomyJan:
However, the command to unlock the map is already exists in [SetPropCommand](https://github.com/Grasscutters/Grasscutter/blob/8b83dbf58c7fdfdd5e1a2f678966a414433e6699/src/main/java/emu/grasscutter/command/commands/SetPropCommand.java#L235), just use `setprop unlockmap <SceneId>` . 
Also, even if you cannot provide translations for all languages, you should add English translations as placeholders for other translation files.

GanyusLeftHorn:
The whole unlocking the map via `/prop` thing was only made because we once had a working implementation for Statues, which would give the player a way to unlock the map naturally. However, that implementation has broken for some unknown reason ...

So, as an alternative to this new command, I would suggest the following two options:
1.) Figure out why Statues are broken, and fix them.
2.) Make the map unlocked by default again (only the map though, not teleport waypoints).

PDMengQAQ:
You can extract this command from [SetPropCommand]

KingRainbow44:

PR #2023, opened 2023-01-17
build: Dockerfile

kilosonc:
Signed-off-by: kiloson <4closetool3@gmail.com>

## Description

Add dockerfile to simplify the deployment.

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


kilosonc:
Hi, @GanyusLeftHorn @KingRainbow44 , could you help me to review this pr?
I thoughts it's good, for it could simplify the deployment so much.

image/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM openjdk:17-jdk-slim
+
+RUN apt update && \
+    apt install -y unzip wget && \
+    mkdir /grasscutter && cd /grasscutter && \
+    wget -O grasscutter.jar https://github.com/Grasscutters/Grasscutter/releases/download/v1.4.3/grasscutter-1.4.3.jar && \
```
KingRainbow44, line 6: Instead of specifically specifying `1.4.3`, add a variable for it? (not completely familiar with `Dockerfile`)

image/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    wget https://raw.githubusercontent.com/Grasscutters/Grasscutter/development/keystore.p12 && \
+    wget -P /grasscutter/resources https://git.crepe.moe/grasscutters/Grasscutter_Resources/-/archive/main/Grasscutter_Resources-main.zip && \
+    unzip -d /grasscutter/resources /grasscutter/resources/Grasscutter_Resources-main.zip && rm /grasscutter/resources/Grasscutter_Resources-main.zip && \
+    unzip -d /grasscutter/resources /grasscutter/resources/Grasscutter_Resources-main/Grasscutter_Resources-3.3.zip && rm -rf /grasscutter/resources/Grasscutter_Resources-main && \
```
KingRainbow44, line 10: Instead of specifically specifying `3.3`, add a variable for it? (not completely familiar with `Dockerfile`)

image/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    mv /grasscutter/resources/Grasscutter_Resources-3.3/Resources/* /grasscutter/resources/ && rm -rf /grasscutter/resources/Grasscutter_Resources-3.3 && \
+    apt remove -y unzip wget
+
+EXPOSE 443
```
KingRainbow44, line 14: These should not be hardcoded values, as they can be configured in the configuration file. (if need be, add something to keep the config consistent, again not familiar with how `Dockerfile` works)

KingRainbow44 (changes requested):

jd-raymaker:
You can use `--build-arg` and the `ARG` instruction.
Example `docker build --build-arg version=1.4.3 .`
Read more about it [here](https://docs.docker.com/engine/reference/builder/#arg)

image/Dockerfile
jd-raymaker: @closetool 
The `EXPOSE` instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.
[Dockerfile reference](https://docs.docker.com/engine/reference/builder/#expose)

I suggest these can be removed as the ports are documented in the wiki and the configuration file

RealistikDash:
Wouldn't it theoretically be better to just use the source code from the repo and build it? This would allow you to
- Copy the files rather than having to download them
- Not have to keep track of version in the dockerfile
- Allow you to run test environments inside docker containers (especially as you get compose involved for things such as MongoDB)
- Is overall better practise

Now that would come with the overhead of setting up a build environment inside docker, but assuming its done properly, docker would be able to cache most of it.

kilosonc:
Sorry for the delay guys
I realized that I did not handle the version of jar and resources.
@RealistikDash You are right, I'll make a Dockerfile to build grasscutter from the source code, but I'm wondering where to find the resources files? Is it also in the source code? Hi @KingRainbow44 @jd-raymaker, could you tell me something about resources? Could I build it from the source code?

KingRainbow44:
resources are fetched from other repositories. this is the primary drawback with a `Dockerfile`: the location of resources is not guaranteed to be in the same place. if anything, this should be the only hardcoded variable, or fetched from system environment?

RealistikDash:
Theoretically, you could assume that the user has manually placed the resources in the correct location (as if they were normally running GC) and copy that in (or possibly even use a volume)?

kilosonc:
Thanks @RealistikDash , but I want to takes resources into image, because it's the advantage of image, containing everything required.
@KingRainbow44 I'm going to make an assumption, there's a variable named `RESOURCE_URL`, pointing to the zipped resources. And I want to add a github action to build the image at every time you tagged a new release, so can you please add RESOURCE_URL, DOCKERHUB_USERNAME and DOCKERHUB_TOKEN to github action secret, what do you think of？

KingRainbow44:
resources **cannot** be built by a GitHub Action worker!

KingRainbow44:
closing due to inactivity; feel free to reopen the PR if you have changes!

PR #2052, opened 2023-02-08
Fix tower guard goal

NotThorny:
## Description

Fixes the goal value for guard levels in the tower, allowing them to finish correctly.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


Hartie95:
Looks good, param4 is probably the moster to kill count, param5 the cfg id of the gadget to protect.
The cfg being ignored could probably lead to problems in the future though, if there is more then one gadget in a group.
I will add your fix to my [quests fork](https://github.com/Anime-Game-Servers/Grasscutter-Quests/tree/quests), where a few more problems with challenges are fixed already (like using the DungeonChallengeConfigData to identify the type of the challenge instead of guessing it)

GanyusLeftHorn (approved):

PR #2133, opened 2023-04-27
fix dragon fight

scooterboo:
## Description

Scene was being updated while it was being deleted. Slapped a synchronized on deregisterScene so that it doesn't run at the same time as Scene.java's checkGroups()

## Issues fixed by this PR
https://cdn.discordapp.com/attachments/1091570006464135273/1100238630011801600/buggeddvalin.mp4
After flying down to the plaza, the dragon fight cutscene would play, but then it would leave you in the overworld where you would eventually disconnect. What was happening is that the main thread was being threadlocked due to Scene.java's checkGroups running at the same time as World.java's deregisterScene

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/world/World.java
+++ b/src/main/java/emu/grasscutter/game/world/World.java
@@ -216,7 +216,7 @@ public class World implements Iterable<Player> {
         this.getScenes().put(scene.getId(), scene);
     }
 
-    public void deregisterScene(Scene scene) {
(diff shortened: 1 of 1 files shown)


scooterboo:
Nope. I take it back. It didn't fix it.

image

PR #2159, opened 2023-05-20
Partially fix Chasing Shadows

scooterboo:
## Description
When you talked to one of the guards, most of the quest line would be marked finished. This was due to a bug in how QUEST_CONTENT_FINISH_PLOT was handled.

In a separate bug, 2010140 was being finished early because quest progress was not being cleared during rewind in clearProgress().

## Issues fixed by this PR
Chasing Shadows works a little bit better in the sense that a car crash is better than a train crash.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/player/PlayerProgress.java
+++ b/src/main/java/emu/grasscutter/game/player/PlayerProgress.java
@@ -78,6 +78,10 @@ public class PlayerProgress {
         return questProgressCountMap.merge(progressId, count, Integer::sum);
     }
 
+    public int resetCurrentProgress(int progressId) {
+        return questProgressCountMap.merge(progressId,0,Integer::min);
+    }
+
     @Entity
     @NoArgsConstructor
     public static class ItemEntry {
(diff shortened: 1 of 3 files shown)


src/main/java/emu/grasscutter/game/player/PlayerProgress.java
```diff
@@ -78,6 +78,10 @@ public int addToCurrentProgress(int progressId, int count) {
         return questProgressCountMap.merge(progressId, count, Integer::sum);
     }
 
+    public int resetCurrentProgress(int progressId) {
+        return questProgressCountMap.merge(progressId,0,Integer::min);
```
KingRainbow44, line 82:
```suggestion
        return questProgressCountMap.merge(progressId, 0, Integer::min);
```

KingRainbow44 (changes requested):

src/main/java/emu/grasscutter/game/player/PlayerProgress.java
KingRainbow44: nvm this will be fixed by spotless anyways

KingRainbow44 (approved):

PR #2161, opened 2023-05-21
Triggers should only trigger triggers they were triggered for.

scooterboo:
## Description
The stuff that caused a trigger should always be looked at when deciding if another quest content should also trigger.
otherwise, quests could trigger again for an unrelated, but same category of trigger.
For example, all quests were checked if they are in their trigger region any time the player enters any region even if it's an unrelated region. This could cause a unrelated event to fire twice.

## Issues fixed by this PR

Unfortunately, I couldn't find any specific bug this fixed. Technically it would have fixed the "Quest {} doesn't have trigger {} registered." spam?

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/quest/content/ContentClearGroupMonster.java
+++ b/src/main/java/emu/grasscutter/game/quest/content/ContentClearGroupMonster.java
@@ -14,7 +14,7 @@ public class ContentClearGroupMonster extends BaseContent {
     public boolean execute(
             GameQuest quest, QuestData.QuestContentCondition condition, String paramStr, int... params) {
         val groupId = condition.getParam()[0];
-
+        if (groupId != params[0]) return false;
         return quest.getOwner().getScene().getScriptManager().isClearedGroupMonsters(groupId);
     }
 }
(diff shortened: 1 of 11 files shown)


Hartie95:
This might fail on quests, where a quest has more then one of the same content handlers.
(see [here](https://github.com/Grasscutters/Grasscutter/blob/e05a29ecf1d89b1c5525fc6302216d97f2ca0b50/src/main/java/emu/grasscutter/game/quest/GameMainQuest.java#L487))

scooterboo:
> This might fail on quests, where a quest has more then one of the same content handlers. (see [here](https://github.com/Grasscutters/Grasscutter/blob/e05a29ecf1d89b1c5525fc6302216d97f2ca0b50/src/main/java/emu/grasscutter/game/quest/GameMainQuest.java#L487))

Is that a my code problem or a handling code in general problem?

Hartie95:
> Is that a my code problem or a handling code in general problem?

Probably a bit on both sides.
Keep in mind, the result of the handlers is used to determine if something is finished, so if you return false in case its not matching, it will handle it as something is not finished, even if it was finished before.
You can either check the params before running the handler code, or you can just make sure quest update stuff only happens on a change (or both).

For the accept condition cache, I used the type+first param+if existing the stringparam as key to identify it, should be ennough here too.

scooterboo:
Ok, don't have time to look into it, but probably line 453 can go below line 454 so that the conditions only get updated if we have a trigger hit. https://github.com/Grasscutters/Grasscutter/blob/e05a29ecf1d89b1c5525fc6302216d97f2ca0b50/src/main/java/emu/grasscutter/game/quest/GameMainQuest.java#L493

scooterboo:
Ok, don't merge this pull just yet, I need a few days to cook.

scooterboo:
I'll come back to this later.

PR #2167, opened 2023-05-26
Fully fix fairy clock for real this time

scooterboo:
## Description
Made the fairy clock accessible after the dragon fight. This also fixes time locking due to quests.
It seems the code had some muddled thinking about how time was locked due to pausing and unpausing vs more permanently locking it due to quests. More booleans were needed

## Issues fixed by this PR
Can access fairy clock later in act 1 of the prologue.
Quests can lock and unlock the time for the duration of the quest.

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

- [x] Lloyd and Muffin helped

### DIFF
--- a/src/main/java/emu/grasscutter/game/world/World.java
+++ b/src/main/java/emu/grasscutter/game/world/World.java
@@ -545,9 +545,9 @@ public final class World implements Iterable<Player> {
      * @param locked True if the world time should be locked.
      */
     public void lockTime(boolean locked) {
-        if(host.getProperty(PlayerProperty.PROP_IS_GAME_TIME_LOCKED) != 0) {
+        if (host.getProperty(PlayerProperty.PROP_IS_GAME_TIME_LOCKED) != 0) {
             this.timeLocked = true;
-        }else{
+        } else {
             this.timeLocked = locked;
         }
         // Notify players of the locking.
--- a/src/main/java/emu/grasscutter/game/world/World.java
+++ b/src/main/java/emu/grasscutter/game/world/World.java
@@ -2,7 +2,6 @@ package emu.grasscutter.game.world;
 
 import static emu.grasscutter.server.event.player.PlayerTeleportEvent.TeleportType.SCRIPT;
 
-import emu.grasscutter.Grasscutter;
 import emu.grasscutter.data.GameData;
 import emu.grasscutter.data.excels.dungeon.DungeonData;
 import emu.grasscutter.game.player.Player;
--- a/src/main/java/emu/grasscutter/game/world/World.java
+++ b/src/main/java/emu/grasscutter/game/world/World.java
@@ -483,8 +483,6 @@ public final class World implements Iterable<Player> {
         // Update the world time.
         this.getWorldTime();
         this.updateTime();
-        // Lock the world time.
-        this.lockTime(paused);
 
         // If the world is being un-paused, update the last update time.
         if (this.isPaused != paused && !paused) {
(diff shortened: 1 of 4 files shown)


Hartie95:
this logic will not allow world to react to it, resulting in time jumps after unlocking the time. So with the current logic lockTime would need to react similar to pause on unlock.

its also probably better to have a default place to call for locking and unlocking, to allow to run the requried logic, so either changing the World.locktime method to set the prop and do the logic, or adding a player.lockTime method to set the prop and notify world if the player is the host might be a good idea.

src/main/java/emu/grasscutter/game/world/World.java
```diff
@@ -545,12 +545,13 @@ public void updateTime() {
      * @param locked True if the world time should be locked.
      */
     public void lockTime(boolean locked) {
-        this.timeLocked = locked;
-
+        if(host.getProperty(PlayerProperty.PROP_IS_GAME_TIME_LOCKED) != 0) {
+            this.timeLocked = true;
+        }else{
+            this.timeLocked = locked;
+        }
```
KingRainbow44: run `gradlew spotlessApply` or
```suggestion
        if (host.getProperty(PlayerProperty.PROP_IS_GAME_TIME_LOCKED) != 0) {
            this.timeLocked = true;
        } else {
            this.timeLocked = locked;
        }
```

src/main/java/emu/grasscutter/server/packet/recv/HandlerClientLockGameTimeNotify.java
```diff
@@ -11,6 +12,6 @@ public final class HandlerClientLockGameTimeNotify extends PacketHandler {
     @Override
     public void handle(GameSession session, byte[] header, byte[] payload) throws Exception {
         var packet = ClientLockGameTimeNotify.parseFrom(payload);
-        session.getPlayer().getWorld().lockTime(packet.getIsLock());
+        session.getPlayer().setProperty(PlayerProperty.PROP_IS_GAME_TIME_LOCKED, packet.getIsLock() ? 1 : 0);
```
KingRainbow44: this should be able to take a boolean, correct me if im wrong

KingRainbow44 (changes requested):

src/main/java/emu/grasscutter/server/packet/recv/HandlerClientLockGameTimeNotify.java
scooterboo: Maybe, but I wanted to play it safe.

src/main/java/emu/grasscutter/game/world/World.java
scooterboo: Ah! So that's where spotlessApply is!

src/main/java/emu/grasscutter/server/packet/recv/HandlerClientLockGameTimeNotify.java
KingRainbow44:
```suggestion
        session.getPlayer().setProperty(PlayerProperty.PROP_IS_GAME_TIME_LOCKED, packet.getIsLock());
```

KingRainbow44 (changes requested):

src/main/java/emu/grasscutter/server/packet/recv/HandlerClientLockGameTimeNotify.java
KingRainbow44: yes, yes it can take a boolean lol

scooterboo:
I have removed all code not related to getting the clock working. Pausing still works as expected, but without any use of time locking in the quests, I'm not testing quest based time locking at this time.

Please argue about specific implementation of how time stopping works in a different PR. This one is about the fairy clock only.

KingRainbow44 (approved):

PR #2176, opened 2023-06-01
Fix issues with regions

scooterboo:
## Description
Luckily, SceneRegion and TriggerExcelConfigData both have group numbers, so we can use those to differentiate regions!

Ice guy being missing in front of his dungeon has plagued us for many moons. 
Today, we finally have a solution.

In TriggerExcelConfigData, some region numbers are repeated. This causes the game to get confused about which region it is entering. We now differentiate them based on group ID as well.

## Issues fixed by this PR
Ice guy shows up!!! No need to relog for the entirety of act 1.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


KingRainbow44:
change branches

PR #2189, opened 2023-06-04
Do not unlock unnecessary scene points during '/prop unlockmap on'

iTruth:
## Description

## Issues fixed by this PR
Do not unlock unnecessary scene points during '/prop unlockmap on'  
the "unnecessary scene points" are:  
1. forbidSimpleUnlock is true
2. scene type is "SceneBuildingPoint" and unlocked is false

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


KingRainbow44:
can you make this a config option?

iTruth:
> can you make this a config option?

certainly, but I prefer add an option for 'unlockmap' because I think for two same abstract machine, if they connect to two different servers, the behavior should be the same when they execute "/prop unlockmap on" even that two server have different config, since a command with same arguments cannot have multiple semantics.  
as you can see, I added 'all' for 'unlockmap', I think this should consistent with semantics, because the value in setProp does not need to have only three states, for example, I can also add 'max' option for 'worldlevel' and 'BP Level', this consistent with semantics too. but obvious, '/prop unlockmap on' does not consistent with semantics but I can't remove this usage because it's too late for now.  
let me know if you have different opinion.  

seems like this pull request is out of order, I'll reopen a new pull request later

PR #2196, opened 2023-06-07
Add option "all" for UnlockMap and change the default behavior for "/prop unlockmap on"

iTruth:
## Description

## Issues fixed by this PR
Add option "all" to unlock all the scene points include unnecessary scene points
Change the behavior for "/prop unlockmap on", now that command will not unlock unnecessary scene points to  be closer to the official server. and I think most people doesn't need these scene points and it will effect gaming experience.
the "unnecessary scene points" are:
1.  forbidSimpleUnlock is true
2. scene type is "SceneBuildingPoint" and unlocked is false

## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/SetPropCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/SetPropCommand.java
@@ -222,6 +222,21 @@ public final class SetPropCommand implements CommandHandler {
         GameData.getScenePointsPerScene()
                 .forEach(
                         (sceneId, scenePoints) -> {
+                            for (var p : scenePoints) {
+                                var scenePointEentry =
+                                    GameData.getScenePointEntryById(sceneId, p);
+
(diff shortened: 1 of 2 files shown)


KingRainbow44:
do the other handlers know how to handle -2?

iTruth:
> do the other handlers know how to handle -2?

yes, but not only -2, all the handlers can handle all the possible numbers properly. but might have different semantics. I don't think we need to check semantics for this, because nobody will execute '/prop bp on' to set his BP_LEVEL to 1. and execute '/prop wl toggle' to trigger a warning. but it's OK if they really do this.
and also, the option 'all' is only for 'UnlockMap' for special usage and I think most people don't need this, so for most people, they can simply ignore that option and use this command just like before.
let me know if you don't think so

PR #2203, opened 2023-06-10
fix description of command `quest` and `list`

TomyJan:
## Description

fix description of command `quest` and `list`

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/QuestCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/QuestCommand.java
@@ -21,7 +21,7 @@ public final class QuestCommand implements CommandHandler {
 
     @Override
     public void execute(Player sender, Player targetPlayer, List<String> args) {
-        if (args.size() != 2  || (args.size() == 1 && !args.get(0).toLowerCase().equals("dungeons")))
+        if (args.size() != 2  || (args.size() == 1 && !args.get(0).toLowerCase().equals("dungeons"))) {
             sendUsageMessage(sender);
             return;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/emu/grasscutter/command/commands/QuestCommand.java
```diff
@@ -11,14 +11,17 @@
...
+            "dungeons"        
+        },
         permission = "player.quest",
         permissionTargeted = "player.quest.others")
 public final class QuestCommand implements CommandHandler {
 
     @Override
     public void execute(Player sender, Player targetPlayer, List<String> args) {
-        if (args.size() != 2) {
+        if (args.size() != 2  || (args.size() == 1 && args.get(0).toLowerCase() != "dungeons")) {
```
der3318: String's `!=` compares the object reference, not the content. Use `equals()` instead.

```suggestion
        if (args.size() != 2  || (args.size() == 1 && !args.get(0).toLowerCase().equals("dungeons")))
```

der3318: BTW, looks like command `/quest dungeons` never works. It's going to hit the exception below (line 33) before entering the switch block:

```java
        try {
            questId = Integer.parseInt(args.get(1));
        } catch (Exception e) {
            CommandHandler.sendMessage(sender, translate(sender, "commands.quest.invalid_id"));
            return;
        }
```

KingRainbow44 (approved):

PR #2208, opened 2023-06-13
add missing routes for CN client & make third party login channel controlled by plugin

TomyJan:
## Description

1. Add missing routes for CN client.
2. make third party login channel controlled by plugin
3. Improve the prompt text for version mismatch.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/quest/content/ContentFinishPlot.java
+++ b/src/main/java/emu/grasscutter/game/quest/content/ContentFinishPlot.java
@@ -1,9 +1,9 @@
 package emu.grasscutter.game.quest.content;
 
+import static emu.grasscutter.game.quest.enums.QuestContent.QUEST_CONTENT_FINISH_PLOT;
+
 import emu.grasscutter.data.excels.quest.QuestData;
 import emu.grasscutter.game.quest.*;
-
-import static emu.grasscutter.game.quest.enums.QuestContent.QUEST_CONTENT_FINISH_PLOT;
 
 @QuestValueContent(QUEST_CONTENT_FINISH_PLOT)
 public class ContentFinishPlot extends BaseContent {
--- a/src/main/java/emu/grasscutter/game/quest/content/ContentNotFinishPlot.java
+++ b/src/main/java/emu/grasscutter/game/quest/content/ContentNotFinishPlot.java
@@ -1,9 +1,9 @@
 package emu.grasscutter.game.quest.content;
 
+import static emu.grasscutter.game.quest.enums.QuestContent.QUEST_CONTENT_NOT_FINISH_PLOT;
+
 import emu.grasscutter.data.excels.quest.QuestData;
 import emu.grasscutter.game.quest.*;
-
-import static emu.grasscutter.game.quest.enums.QuestContent.QUEST_CONTENT_NOT_FINISH_PLOT;
 
 @QuestValueContent(QUEST_CONTENT_NOT_FINISH_PLOT)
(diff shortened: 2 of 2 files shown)


Hartie95:
Won't disabling third party login remove the ability to use plugins for custom oauth login with something like these?
https://github.com/Xtao-Labs/GCAuth-OAuth
https://github.com/YukariChiba/GCAuth-OpenID

TomyJan:
> Won't disabling third party login remove the ability to use plugins for custom oauth login with something like these? https://github.com/Xtao-Labs/GCAuth-OAuth https://github.com/YukariChiba/GCAuth-OpenID

Yes, the login panel will not have any thirdparty icon.
But these plugins are no longer updated, would no longer be compatible.
Or should I write it to be configurable?

Hartie95:
> > Won't disabling third party login remove the ability to use plugins for custom oauth login with something like these? https://github.com/Xtao-Labs/GCAuth-OAuth https://github.com/YukariChiba/GCAuth-OpenID
> 
> Yes, the login panel will not have any thirdparty icon. But these plugins are no longer updated, would no longer be compatible. Or should I write it to be configurable?

I for example use the openId plugin, so at some point I will probably pick up the openId plugin, if no one else does it.
But even without it, at least giving the option to enable them would be good, in case some one else want to build something like that, for example for discord based auth.
Could either be a direct config, or maybe something that plugins can register at, which then automatically enables them. The later would probably be more user friendly, though it would need to be documented somewhere for plugin devs.

TomyJan:
> > > Won't disabling third party login remove the ability to use plugins for custom oauth login with something like these? https://github.com/Xtao-Labs/GCAuth-OAuth https://github.com/YukariChiba/GCAuth-OpenID
> > 
> > 
> > Yes, the login panel will not have any thirdparty icon. But these plugins are no longer updated, would no longer be compatible. Or should I write it to be configurable?
> 
> I for example use the openId plugin, so at some point I will probably pick up the openId plugin, if no one else does it. But even without it, at least giving the option to enable them would be good, in case some one else want to build something like that, for example for discord based auth. Could either be a direct config, or maybe something that plugins can register at, which then automatically enables them. The later would probably be more user friendly, though it would need to be documented somewhere for plugin devs.

Thanks for your advice. I couldnt find a good way to control it by plugin, so instead, I will work on providing a configuration option to config avaible thirdparty login channel.

KingRainbow44 (changes requested):
please make third party login configurable, or check if the oauth authenticator in the authentication system isn’t null/default.

KingRainbow44:
additionally if this is a WIP PR, mark it as draft.

other note some people use cn QR code login, keep that with the authenticators as well, or don’t remove it outright

TomyJan:
All done, but in fact gc does not currently have a mechanism to refuse to load a plugin, which is what I and  #2206  need, so anyone going to implement it?

KingRainbow44 (changes requested):
change your string concatenation to building a JSON object
use `List` over `String[]`
move login channels away from the plugin config, add it to the server handle
move the third party login channel array away from the plugin manager, add it to the authentication system
change `version_outdate` to `version_outdated`
you can use `""" """` and `%s` for the smaller JSON string inserts

KingRainbow44:
not completely sure why this was merged, as I didn’t merge it

TomyJan:
> not completely sure why this was merged, as I didn’t merge it

idk, I'm busy these days, I'll open a new pr to fix it as soon as I have time.

PR #2218, opened 2023-06-18
Fix Librarian Story Quest

scooterboo:
## Description
Various fixes to make Librarian's story quest go from red to dark green.

4 in 1 pack!

## Issues fixed by this PR
GadgetGatherObject.java ->
When gathering an object, send a EventType.EVENT_GATHER to ScriptArgs so that those kinds of triggers will happen.
Quest 1010115 was not finishing because the lua in scene 1004 group 201004003 was not being told that the book was taken from the shelf.

ScriptLib.java -> config_id
There was a CreateGadget with config_id of 0 in scene 20010, group 220010006. I don't know what the crap it is supposed to do, but it keeps the group from loading.
I made it check for config_id of 0 and return a 0 in that case.

ScriptLib.java -> VisionType.VISION_TYPE_REMOVE
When we REMOVE an entity, we should send a VisionType.VISION_TYPE_REMOVE not a VisionType.VISION_TYPE_DIE
In the dungeon in this quest, there's an invisible floor that keeps you from taking the platform down. With VisionType.VISION_TYPE_DIE the invisible floor remained in place and with VisionType.VISION_TYPE_REMOVE it was removed.

GameEntity.java, EntityIdType.java, EntityType.java ->
This is the only fix that you might want to look into.
in the lua scripts, the values of EntityType are used for things like EntityType.GADGET . For Grasscutter, the values from EntityIdType are used to make unique IDs. I could not find a way to get an entity's EntityType from a GameEntity. So instead, I put a map in EntityIdType that converts EntityIdTypes to EntityType. I checked GameEntity's getEntityType() and it is only used when EntityType is needed, but it returns EntityIdTypes instead. I added Region to EntityType because it's used in the lua in the electric islands (might as well add it now).
In the game, when trying to remove that invisible floor I was talking about earlier, it would return a 1 instead of a 0 in ScriptLib.java's RemoveEntityByConfigId (scene 20010 group 220010010) and therefor fail. This was because entity.getEntityType() (which returns a EntityIdType) was being checked against  entityType (which was a EntityType). This made it so that the final worktop in the dungeon would appear to not do anything when you interacted with it.

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/entity/GameEntity.java
+++ b/src/main/java/emu/grasscutter/game/entity/GameEntity.java
@@ -53,7 +53,7 @@ public abstract class GameEntity {
     public abstract void initAbilities();
 
     public int getEntityType() {
-        return this.getId() >> 24;
+        return EntityIdType.toEntityType(this.getId() >> 24);
     }
 
     public abstract int getEntityTypeId();
--- a/src/main/java/emu/grasscutter/game/entity/GameEntity.java
+++ b/src/main/java/emu/grasscutter/game/entity/GameEntity.java
@@ -52,7 +52,7 @@ public abstract class GameEntity {
 
     public abstract void initAbilities();
 
-    public int getEntityType() {
+    public EntityType getEntityType() {
         return EntityIdType.toEntityType(this.getId() >> 24);
     }
 
--- a/src/main/java/emu/grasscutter/game/props/EntityType.java
+++ b/src/main/java/emu/grasscutter/game/props/EntityType.java
@@ -75,7 +75,7 @@ public enum EntityType implements IntValueEnum {
     Screen(64),
     EchoShell(65),
     UIInteractGadget(66),
-    Region(67),
+    Region(98),
     PlaceHolder(99);
 
     private static final Int2ObjectMap<EntityType> map = new Int2ObjectOpenHashMap<>();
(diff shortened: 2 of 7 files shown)


src/main/java/emu/grasscutter/game/props/EntityType.java
```diff
@@ -75,6 +75,7 @@ public enum EntityType implements IntValueEnum {
     Screen(64),
     EchoShell(65),
     UIInteractGadget(66),
+    Region(67),
```
Hartie95: Best use some high value for custom stuff, since otherwise we will quickly run into collisions.
Also best add a comment that this is nothing real from official, and needs more investigation.

src/main/java/emu/grasscutter/game/entity/GameEntity.java
```diff
@@ -53,7 +53,7 @@ public GameEntity(Scene scene) {
     public abstract void initAbilities();
 
     public int getEntityType() {
-        return this.getId() >> 24;
+        return EntityIdType.toEntityType(this.getId() >> 24);
```
Hartie95, line 56: Wouldn't it be better to actually return the enum type, if you do the conversion? Otherwise something might expect the wrong value again.

Hartie95 (approved):

KingRainbow44 (approved):

PR #2220, opened 2023-06-18
Use the old version of the drop system.

ERthereal:
## Description

## Issues fixed by this PR
In the new version of Grasscutter, monsters do not drop any items

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] My pull request is unique and no other pull requests have been opened for these changes

### DIFF
--- a/src/main/java/emu/grasscutter/game/world/Scene.java
+++ b/src/main/java/emu/grasscutter/game/world/Scene.java
@@ -496,7 +496,7 @@ public final class Scene {
         // Reward drop
         var world = this.getWorld();
         if (target instanceof EntityMonster && this.getSceneType() != SceneType.SCENE_DUNGEON) {
-            world.getServer().getDropSystem().callDrop((EntityMonster) target);
+            world.getServer().getDropSystemLegacy().callDrop((EntityMonster) target);
         }
 
         // Remove entity from world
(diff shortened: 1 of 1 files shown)


Hartie95:
This looks kinda like a workaround, instead of addressing the problem. 
Did you make sure your resources are up to date? Did you debug by chance why the fallback to the old drop system doesn't happen?

Hartie95:
Oh, do you by chance have scripts disabled for big world in the config? This would cause not meta object to exist, and thus not create drops.
But in that's case, you should check if meta == null || !handleDropMonster

scooterboo:
Using /troubleshoot can figure this stuff out pretty quickly.

ERthereal:
> 这看起来有点像解决方法，而不是解决问题。您是否确保您的资源是最新的？您是否偶然调试了为什么不会回退到旧的丢弃系统？

I confirm that my resources are up-to-date. When I start Grasscutter with debug mode, I found that if the JSON related to dropping does not exist in the resources, Grasscutter can still load the JSON in the resources successfully. This should mean that Grasscutter will never return to the old version of the drop system.

ERthereal:
> 哦，您是否偶然在配置中为大世界禁用了脚本？这将导致元对象不存在，因此不会创建丢弃。但在这种情况下，您应该检查 meta == 空 ||！handleDropMonster

It seems unrelated to whether ScriptInBigWorld is enabled or not.

PR #2225, opened 2023-06-22
Fix ore not falling from rocks

scooterboo:
## Description
Fix ore not falling from rocks.

Note however that the client tells the server to remove the rock, and then tells the server to do the drops.
It is not safe to call player.getScene().getEntityById().

So I implemented a cache system for recently deleted entities.
I've set the size limit to an arbitrary 500 entities. That's about 2 loading zones full of entities. 
removeEldestEntry comes in clutch. If it returns true, it deletes the oldest entry when .put() is used.

## Issues fixed by this PR
Ores did not drop.

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/ability/AbilityManager.java
+++ b/src/main/java/emu/grasscutter/game/ability/AbilityManager.java
@@ -9,9 +9,9 @@ import emu.grasscutter.game.ability.actions.*;
 import emu.grasscutter.game.ability.mixins.*;
 import emu.grasscutter.game.entity.EntityGadget;
 import emu.grasscutter.game.entity.GameEntity;
+import emu.grasscutter.game.entity.gadget.GadgetGatherObject;
 import emu.grasscutter.game.player.*;
 import emu.grasscutter.game.props.FightProperty;
-import emu.grasscutter.game.entity.gadget.GadgetGatherObject;
 import emu.grasscutter.net.proto.AbilityInvokeEntryOuterClass.AbilityInvokeEntry;
 import emu.grasscutter.net.proto.AbilityMetaAddAbilityOuterClass.AbilityMetaAddAbility;
 import emu.grasscutter.net.proto.AbilityMetaModifierChangeOuterClass.AbilityMetaModifierChange;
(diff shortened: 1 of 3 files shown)


src/main/java/emu/grasscutter/game/ability/AbilityManager.java
```diff
@@ -364,6 +366,23 @@ private void handleModifierChange(AbilityInvokeEntry invoke) throws Exception {
...
+        if (entity == null) {
+            Grasscutter.getLogger().debug("Entity not found: {}", invoke.getEntityId());
+            entity = this.player.getScene().getRemovedEntityById(invoke.getEntityId());
+            if (entity == null) return;
+            Grasscutter.getLogger().debug("Using removed Entity: {}", invoke.getEntityId());
+        }
+
+        // Destroying rocks
+        if (entity instanceof EntityGadget targetGadget
+            && targetGadget.getContent() instanceof GadgetGatherObject gatherObject) {
```
Hartie95, line 379: Hmm, looks kinda weird to put this hard coded in here, since its not directly part of the invoke entry actions.
Maybe it would be better to put this as a die action if there is no gadget lua or ability for this.
Might maybe be good to talk with @StartForKiller about this.

Hartie95: applying this to the resources should properly fix the problem with the drops:
https://github.com/Anime-Game-Servers/CustomGCResources/pull/7

src/main/java/emu/grasscutter/game/world/Scene.java
```diff
@@ -75,6 +76,12 @@ public Scene(World world, SceneData sceneData) {
         this.sceneData = sceneData;
         this.players = new CopyOnWriteArrayList<>();
         this.entities = new ConcurrentHashMap<>();
+        this.removedEntities = new LinkedHashMap<>(){
```
StartForKiller, line 79: Please, try not using this types of hacks, it's obvious that this is wrong, so don't use it, try to figure what is the actual problem and not using and old removed code

scooterboo: Sorry. I checked hartie's and I checked your main branches and saw the code still there, so I figured the code was accidentally deleted and needed to be reinstated.

scooterboo:
This will be fixed by https://github.com/Anime-Game-Servers/CustomGCResources/commit/48451327bef84301552e33973d4c0aa8a65c7f71 instead

PR #2226, opened 2023-06-23
Fix unlogging quests

scooterboo:
## Description
I couldn't use /q dubug to turn off debugging on a quest.
loggedQuests.remove() here is getting confused between 

public E remove(int index)
and
public boolean remove(Object o)

because questId is an int.

## Issues fixed by this PR
image

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


scooterboo:
Alternatively, 
loggedQuests.remove((Integer) questId);

PR #2240, opened 2023-06-27
Update region logic

scooterboo:
## Description
Sometimes, there's triggers with actions, but no conditions. It's more accurate to do matching based off of name rather than condition.

## Issues fixed by this PR
Unfortunately, I've only seen this fix the slimes at group 133104058 at 629.781 200.000 573.870
That group is a pretty good example of this happening, though.

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


eternalcomet:
I think this is not a proper fix. We shouldn't treat EVENT_ENTER_REGION differently. The `condition_EVENT_ENTER_REGION_xxx` is checked by lua scripts rather than java side. So we need to add missing ScriptLib functions like `GetRegionConfigId` to make these triggers triggered more naturally. However, sometimes the condition is empty in some triggers, not only for EVENT_ENTER_REGION. So we need to find a general way to handle the event trigger when the condition is empty for other eventTypes, too.

PR #2270, opened 2023-07-20
various small bug fixes

scooterboo:
## Description
Please review carefully!

A few small patches cluttering up my local distro.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
+++ b/src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
@@ -215,7 +215,13 @@ public class SceneScriptManager {
         var suiteData = group.getSuiteByIndex(suiteIndex);
         if (suiteData == null) {
             Grasscutter.getLogger().warn("Group {} suite {} not found", group.id, suiteIndex);
-            return 0;
+            group.setLoaded(false);
+            group.load(this.scene.getId());
+            suiteData = group.getSuiteByIndex(suiteIndex);
+            if (suiteData == null) {
+                return 0;
+            }
+            Grasscutter.getLogger().warn("Group {} suite {} nvm, I found it.", group.id, suiteIndex);
         }
 
         int prevSuiteIndex = groupInstance.getActiveSuiteId();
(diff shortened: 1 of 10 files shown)


src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
```diff
@@ -215,7 +215,13 @@ public int refreshGroup(
         var suiteData = group.getSuiteByIndex(suiteIndex);
         if (suiteData == null) {
             Grasscutter.getLogger().warn("Group {} suite {} not found", group.id, suiteIndex);
-            return 0;
+            group.setLoaded(false);
```
Hartie95, line 218: Whats the benefit of loading it here again? if group.loaded was true, the next try to load it would pretty much result in the same result as the first time, since the loading code never changed.
also this will trigger a reload of the group every time you request an invalid suite index, not sure thats a good idea.

scooterboo: I've been recently getting groups that have null suites here, but the files actually have suite data. I've been unable to locate the root cause and it worries me. Rereading the group does the trick, though. Otherwise, null suites are typically caused by the group having a suite_disk instead of suites.

Hartie95: Hmm, that either means something weird is happening in terms of thread handling, something with the loading logic is very broken, or something overwrites data.
Since otherwise the same load should not result in different result.
Also are you sure the group was actually loaded in the first place?

Because this looks like a really hacky workaround for a bigger problem somewhere else, might create problems in other areas when the lua groups become user session independend.

scooterboo: Ya! Everything else was fine about the group, it had its gadgets and everything, but the suites were null. no suites. I have a suspicion that the rollback plugin or maybe dynamic groups might be causing it, but I don't have any good leads currently.

Hartie95 (approved):

scooterboo:
The changes to ContentQuestStateEqual fix the finish condition handling of quest 202304

PR #2295, opened 2023-08-18
Update RegionHandler.java

CerisWhite:
## Description
This is a small gripe, and you're free to reject this of course, but it seems pointless to regex the string purely to remove the version code when substring is more than sufficient. This also allows all three platform names to be 8 characters even.


Catta1997:
I think that regex is better since you do not know how long version number is

CerisWhite:
This particular use of regex is to remove only the version code (e.g. "3.7") from the string, in order to get the client type (e.g. "OSRELWin", "OSRELiOS", etc), so the length of the actual version code has no bearing. We only need the first 8 characters to identify the client, and regex on the entire string to remove only the version code is more intensive than just taking the first 8 characters.

PR #2306, opened 2023-08-24
Fix errors in chasm

scooterboo:
## Description
Various fixes for making the chasm questing go smoother.

Scene.java:
I have never seen a dynamic group get loaded. It seems that getScriptManager().getGroupInstanceById always returns true.
So instead, I removed the check for it so it always reloads the dynamic group if a quest calls for it.

SceneScriptManager.java cachedSceneGroupsInstances.put:
I was getting all kinds of nasty errors about missing lua files. I tracked it back to these cachedSceneGroupsInstances.put's that were not setting the lua scripts when caching groups. This probably fixes the missing suites problem we had earlier.

SceneScriptManager.java enterEntities and leaveEntities not being 19:
19 are gadgets. Some time ago we changed the region code so that it gets triggered by everything: monsters, avatars, gadgets, you name it. There were some problems where regions were being triggered early due to the region containing a gadget and there being no other checks in the lua to prevent the region from triggering. Form what we've seen, there's nowhere in the code specifically looking for gadgets triggering regions.

SceneScriptManager.java t.getName().length() <= 12:
This is prep for later when we add "require" scripts (require is the lua equivalent of C's #include). In those files, region triggers don't have numbers, so this new code keeps it from erroring out.

ScriptLib.java RefreshGroup:
RefreshGroup is slightly different than official. in official, it seems that RefreshGroup does not cause createEntity to fire, but in grasscutter, it simply kills and creates all the entities. This causes bugs, and I want to track them by changing debug to info.

ScriptLib.java GetOpeningDungeonListByRosterId ModifyClimatePolygonParamTable SetEnvironmentEffectState GetRegionConfigId:
New functions, 'nuff said.

## Issues fixed by this PR
Scene.java: 
When you first go down into the chasm, there's three diary pages you need to look for. They were not being loaded because they were a dynamic group.

SceneScriptManager.java cachedSceneGroupsInstances.put:
big floods of errors.

SceneScriptManager.java enterEntities and leaveEntities not being 19:
emergency food was getting pretty chatty and things were being triggered at distance due to the region containing gadgets. Specifically a big one was this parkour route that takes you up a rope bridge that was being triggered because the marker you were supposed to get close to was a gadget.

SceneScriptManager.java t.getName().length() <= 12:
Errors when adding the require code in later.

ScriptLib.java RefreshGroup:
There's an infinite loop involving RefreshGroup, and this will make it more clear.

ScriptLib.java GetOpeningDungeonListByRosterId ModifyClimatePolygonParamTable SetEnvironmentEffectState GetRegionConfigId:
New functions, 'nuff said.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/world/Scene.java
+++ b/src/main/java/emu/grasscutter/game/world/Scene.java
@@ -910,8 +910,8 @@ public final class Scene {
 
     public int loadDynamicGroup(int group_id) {
         SceneGroup group = getScriptManager().getGroupById(group_id);
-        if (group == null || getScriptManager().getGroupInstanceById(group_id) != null)
-            return -1; // Group not found or already instanced
+        if (group == null)
+            return -1; // Group not found
 
         this.onLoadGroup(new ArrayList<>(List.of(group)));
 
--- a/src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
+++ b/src/main/java/emu/grasscutter/scripts/SceneScriptManager.java
@@ -395,7 +395,11 @@ public class SceneScriptManager {
         var instance = cachedSceneGroupsInstances.getOrDefault(groupId, null);
         if (instance == null) {
             instance = DatabaseHelper.loadGroupInstance(groupId, scene.getWorld().getHost());
-            if (instance != null) cachedSceneGroupsInstances.put(groupId, instance);
+            if (instance != null){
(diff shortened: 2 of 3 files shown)


scooterboo:
Retracting this. These were already committed in dev 4.0

PR #2314, opened 2023-08-27
Bug fixes

NotThorny:
## Description

Fixes some bugs:
- Fix loading with no skill data on traveler after first login (previously used pyro skill data)
- Fix group and sound commands overriding common aliases for give and spawn
- Fix discord link in version mismatch popup
- Update home resource loading

## Issues fixed by this PR

Those about not being able to attack, in support.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:


src/main/java/emu/grasscutter/server/packet/recv/HandlerSetPlayerBornDataReq.java
```diff
@@ -28,10 +28,13 @@ public void handle(GameSession session, byte[] header, byte[] payload) throws Ex
         // Sanity checks
         int avatarId = req.getAvatarId();
         int startingSkillDepot;
+        int elementlessSkillDepot;
```
Hartie95: This should not need hardcoding, since the skillDepotId for both travelers is already managing this via AvatarExcelConfigData.

Hartie95: Handling for non questing could probably also be improved here by setting the skill depot for starting via element, instead of hardcoding the ids, though thats not something you touched in this pr.

NotThorny: Yeah originally I had this here since I noticed skill depot was being set to pyro, but it was resolved in another file - this was partially a remnant of that, and isn't needed.

PR #2333, opened 2023-09-02
9461618188

2y8e9h22:


KingRainbow44:
what does removing the `this.setDispatchKey(true);` do?

KingRainbow44:
you should not be removing a bunch of protos. feel free to reopen this pr if u revert the change

PR #2348, opened 2023-09-08
Implement RegionShape.POLYGON and RegionShape.CYLINDER

scooterboo:
## Description
Kinda surprised this didn't happen sooner, but CYLINDER and POLYGON are valid region types that were never implemented.

I implemented them and took the opportunity to sort them in order and use multiplication instead of pow. (slight speed up)

## Issues fixed by this PR
POLYGON is used in one quest near the chasm, and CYLINDER is used on all those dungeon waypoints that wouldn't move when you got near them.

## Type of changes

- [x] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


KingRainbow44 (approved):
nazrin back at it again with more math

PR #2361, opened 2023-09-14
Better SceneTags

NotThorny:
## Description

Better handling and implementations for scene tags.

- Adds LUA handling of scenetags
- De-hardcodes scenetags so that they can be properly added and removed
- Add command to manually add and remove scenetags

## Type of changes

- [ ] Bug fix
- [x] New feature 
- [x] Enhancement
- [ ] Documentation

## Checklist:


src/main/java/emu/grasscutter/scripts/ScriptLib.java
```diff
@@ -1013,22 +1013,38 @@ public int ChangeToTargetLevelTag(int var1){
...
-        logger.warn("[LUA] Call unimplemented DelSceneTag with {}, {}", sceneId, sceneTagId);
-        //TODO implement
+        logger.debug("[LUA] Call DelSceneTag with {}, {}", sceneId, sceneTagId);
+        var sceneTags = sceneScriptManager.get().getScene().getHost().getSceneTags();
+        // Ensure key exists for given scene
+        if (sceneTags.get(sceneId) == null) {
+            // Can't delete something that doesn't exist
+            return 0;
+        }
+        sceneScriptManager.get().getScene().getHost().getSceneTags().get(sceneId).remove(sceneTagId);
```
Hartie95: There are a lot of things managing scenetags, so it might be good to move the general add/remove logic to the player progress Manager.
Im also relatively sure, we should send some type of package when a tag gets added/removed, to notify the client about it.

NotThorny: I put the packet in the command and totally spaced adding it to the lua handling as well.

Hartie95: Yep, thats why shared code for adding/removing them, for example in player progress Manager, is so helpfull/important ^^

src/main/java/emu/grasscutter/game/player/PlayerProgressManager.java
```diff
@@ -313,4 +315,32 @@ public void addItemObtainedHistory(int id, int count) {
...
     }
+
+    /******************************************************************************************************************
+     ******************************************************************************************************************
+     * SCENETAGS
+     ******************************************************************************************************************
+     *****************************************************************************************************************/
+    public void addSceneTag(int sceneId, int sceneTagId) {
+        // Sanity check
+        if (player.getSceneTags().get(sceneId) == null) {
```
Hartie95: Only a small thing, but you can also do something like this to add the hashset when its not there yet:
```java
player.getSceneTags().computeIfAbsent(sceneId, (key) -> new HashSet<>())
    .add(sceneTagID)
```

Hartie95 (approved):

src/main/java/emu/grasscutter/server/packet/send/PacketPlayerWorldSceneInfoListNotify.java
```diff
@@ -1,46 +1,46 @@
...
         PlayerWorldSceneInfoListNotify.Builder proto =
                 PlayerWorldSceneInfoListNotify.newBuilder()
-                        .addInfoList(PlayerWorldSceneInfo.newBuilder().setSceneId(1).setIsLocked(false).build())
-                        .addInfoList(
-                                PlayerWorldSceneInfo.newBuilder()
-                                        .setSceneId(3)
+                        .addInfoList(PlayerWorldSceneInfo.newBuilder().setSceneId(1).setIsLocked(false).build());
+
+                        // Iterate over all scenes the user has tags for
+                        for (int scene : sceneTags.keySet()) {
```
Hartie95: might not be a good idea to iterate over the scenetags, a scene might not have scenetags, but still need info that needs to be send here.
We probably want to iterate over GameData.getSceneDataMap() filtered by type SCENE_WORLD instead and then get the data we need for them (like the sceneTAgs)

src/main/java/emu/grasscutter/game/player/PlayerProgressManager.java
NotThorny: Ah of course, I was trying to recall the better inline method. That's what I get for making this right before bed.

src/main/java/emu/grasscutter/server/packet/send/PacketPlayerWorldSceneInfoListNotify.java
Hartie95: or maybe in the long term something scene specific that also stores unlock/lock could be good instead

NotThorny: Some dungeons also use scenetags (Dream dungeons, just as an example) so I wouldn't want to limit it quite that much to just scene_world. A full scene data store implementation might be a bit beyond what I want to add (scenetags vs full scene data), though it would make sense if there's going to be more scene info outside of scenetags and map layers eventually.

src/main/java/emu/grasscutter/command/commands/SetSceneTagCommand.java
```diff
@@ -0,0 +1,98 @@
...
+
+        // Remove default SceneTags, as most are "before" or "locked" states
+        allData.stream().filter(sceneTag -> sceneTag.isDefaultValid())
+                // Only remove for big world as some other scenes only have defaults
+                .filter(sceneTag -> sceneTag.getSceneId() == 3)
+                .forEach(sceneTag -> {
+                    targetPlayer.getSceneTags().get(sceneTag.getSceneId()).remove(sceneTag.getId());
+                });
+
+        setSceneTags(targetPlayer);
```
KingRainbow44:
```suggestion
        this.setSceneTags(targetPlayer);
```

src/main/java/emu/grasscutter/command/commands/SetSceneTagCommand.java
```diff
@@ -0,0 +1,98 @@
...
+
+    }
+
+    private void addSceneTag(Player targetPlayer, int scene, int value) {
+        targetPlayer.getProgressManager().addSceneTag(scene, value);
+    }
+
+    private void removeSceneTag(Player targetPlayer, int scene, int value) {
+        targetPlayer.getProgressManager().delSceneTag(scene, value);
+
```
KingRainbow44:
```suggestion
```

src/main/java/emu/grasscutter/command/commands/SetSceneTagCommand.java
```diff
@@ -0,0 +1,98 @@
...
+            if (actionStr.equals("unlockall")) {
+                unlockAllSceneTags(targetPlayer);
+                return;
+            } else {
+                CommandHandler.sendTranslatedMessage(sender, "commands.execution.argument_error");
+                return;
+            }
+        }
+
+        final int userVal = value;
```
KingRainbow44: consider using `lombok.val`

src/main/java/emu/grasscutter/command/commands/SetSceneTagCommand.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public void execute(Player sender, Player targetPlayer, List<String> args) {
+        if (args.size() == 0) {
+            sendUsageMessage(sender);
+            return;
+        }
+
+        String actionStr = args.get(0).toLowerCase();
+        int value = -1;
```
KingRainbow44:
```suggestion
        var actionStr = args.get(0).toLowerCase();
        var value = -1;
```

src/main/java/emu/grasscutter/scripts/ScriptLib.java
```diff
@@ -1013,22 +1013,22 @@ public int ChangeToTargetLevelTag(int var1){
     }
 
     public int AddSceneTag(int sceneId, int sceneTagId){
-        logger.warn("[LUA] Call unimplemented AddSceneTag with {}, {}", sceneId, sceneTagId);
-        //TODO implement
+        logger.debug("[LUA] Call AddSceneTag with {}, {}", sceneId, sceneTagId);
+        sceneScriptManager.get().getScene().getHost().getProgressManager().addSceneTag(sceneId, sceneTagId);
```
KingRainbow44: there is probably a better way of getting the scene or the host

KingRainbow44 (changes requested):

src/main/java/emu/grasscutter/command/commands/SetSceneTagCommand.java
Hartie95: at least actionStr doesn't change, so `val` would be even better for it ^^

src/main/java/emu/grasscutter/scripts/ScriptLib.java
NotThorny: I looked at other methods but it seemed like this was going to be the cleanest method from within ScriptLib. If you'd prefer new imports & methods I'm open to suggestion.

KingRainbow44 (approved):

PR #2368, opened 2023-09-16
Fix freeze on startup

jie65535:
## Description

On some devices, after starting the server and displaying ‘Loading Grasscutter’, there is no further action. The CPU usage is shown as 0% in the task manager.

By adding logs, I found that `ResourceLoader.loadResources()` will be executed during static loading, and all resources will be loaded simultaneously using `parallel()`.

The specific calling path is: `Grasscutter static init` -> `Utils.startupCheck()` -> `DataLoader.checkAllFiles()` -> `generateGachaMappings()` -> `Tools.createGachaMappings(path)` -> `ResourceLoader.loadResources()` -> `parallel()` -> Freeze.

Freeze screenshot

If parallelism is removed,  it can be loaded normally, so I wonder if the thread pool has not been started during static initialization? Causes the parallel method to not continue?

Sync load

Therefore, I adjusted the loading order, placed the generation logic after resource loading, and deleted the resource loading step during generation.

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -158,7 +158,7 @@ public final class Grasscutter {
 
             // Generate handbooks.
             Tools.createGmHandbooks(false);
-
+            // Generate gacha mappings.
             Tools.generateGachaMappings();
         }
 
--- a/src/main/java/emu/grasscutter/Grasscutter.java
+++ b/src/main/java/emu/grasscutter/Grasscutter.java
@@ -158,6 +158,8 @@ public final class Grasscutter {
 
             // Generate handbooks.
             Tools.createGmHandbooks(false);
+
+            Tools.generateGachaMappings();
         }
 
         // Start servers.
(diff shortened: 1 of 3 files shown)


src/main/java/emu/grasscutter/Grasscutter.java
```diff
@@ -158,6 +158,8 @@ public static void main(String[] args) throws Exception {
 
             // Generate handbooks.
             Tools.createGmHandbooks(false);
+
+            Tools.generateGachaMappings();
```
KingRainbow44:
```suggestion
            // Generate gacha mappings.
            Tools.generateGachaMappings();
```

src/main/java/emu/grasscutter/tools/Tools.java
```diff
@@ -311,8 +312,20 @@ public static List<String> createGachaMappingJsons() {
...
+                Tools.createGachaMappings(path);
+            } catch (Exception exception) {
+                Grasscutter.getLogger().warn("Failed to create gacha mappings. \n" + exception);
+            }
+        }
+    }
+
     public static void createGachaMappings(Path location) throws IOException {
-        ResourceLoader.loadResources();
+        // ResourceLoader.loadResources();
```
KingRainbow44:
```suggestion
```

KingRainbow44 (changes requested):

KingRainbow44 (approved):

PR #2395, opened 2023-10-05
Alphabetize and Format ScriptLib.java

scooterboo:
## Description
I deeply apologize to anyone who is trying to find the history of the code from before today.

- [ ] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [x] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/scripts/ScriptLib.java
+++ b/src/main/java/emu/grasscutter/scripts/ScriptLib.java
@@ -44,11 +44,11 @@ public class ScriptLib {
         this.currentEntity = new FastThreadLocal<>();
     }
 
-    public void setSceneScriptManager(SceneScriptManager sceneScriptManager){
+    public void setSceneScriptManager(SceneScriptManager sceneScriptManager) {
         this.sceneScriptManager.set(sceneScriptManager);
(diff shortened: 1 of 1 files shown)


Hartie95:
Might be more of a preferences things, but isn't it better in the long run to have them grouped by their function? 
Otherwise you will have for example all variable handling on completely different parts of the file, making it hard to follow usage.

scooterboo:
I tried doing that but gave up on it. There's just too many functions.

There would be trouble where it would be hard for new people to follow a sorting that like. If a new function was made, it'd be very likely that someone would insert it into the wrong place.

Better would be if we somehow could manage to split scriptlib into multiple files.

eternalcomet:
Modern IDE like vscode and intellij idea are able to show the structure or outline of the file, which make it easier to find specific method. Besides, as you mentioned above, such refactor make it almost impossible to find code history.

KingRainbow44 (approved):

KingRainbow44:
> such refactor make it almost impossible to find code history

intellij idea's git viewer allows you to see the commit history of a file, including who changed it. this shouldn't affect anything.

Issue #2282, opened 2023-07-30
[Bug] No response was received after interacting with the roots after completing the Domain of Blessing


Phong940253:
**Did you look for other closed issues that have the same problem?**

Yes

**Describe the bug**

No response was received after interacting with the roots after completing the Domain of Blessing.

**Which branch did you use?**

Grasscutter version: 1.6.2-76e940bc

**Screenshots**

![ảnh](https://github.com/Grasscutters/Grasscutter/assets/40328498/567f3c6a-5a48-4f1f-9a98-b42f870b8d3a)

After the interaction no response occurs
![ảnh](https://github.com/Grasscutters/Grasscutter/assets/40328498/7b790bcf-9262-4658-a6e4-f98e517a21c8)

**Additional context**

Curr-Ronaldo:
请问有res资源文件嘛

PR #2398, opened 2023-10-10
Fix daily dungeon flow

longfruit:
Fix daily dungeon flow

## Description

* Fixed "Locked" dungeon entry issue. How this works:
    * Dungeons start out Locked on client side. Interacting with dungeons in this state causes the client to send GetDungeonEntryExploreConditionReq to get level and/or quest completion requirement for dungeon entry.
    * The server would respond with GetDungeonEntryExploreConditionRsp to show the conditions.
        * On player level (AR) up, the server _should_ send DungeonEntryToBeExploreNotify to tell the client some dungeon is ready to be explored. This gives that dungeon the "Unknown" text bubble in the client world map.
        * In this state, the client will no longer send GetDungeonEntryExploreConditionReq and instead send DungeonEntryInfoReq to enumerate dungeon entries as usual, as well as the packet to unlock the teleport waypoint.
    * Unfortunately this fix still needs improvement: I don't know how to make the unlock state persistent.
        * On relog, the dungeon goes back to its locked state. However, its teleport waypoint is still active.
            * In this state, player can still teleport to it, but on interaction, client resumes sending GetDungeonEntryExploreConditionReq.
            * The server has to resend the DungeonEntryToBeExploreNotify again to put it in the unlock state for client to enter.
* Fix daily dungeon enumeration issue.
    * The "dungeonRandomList" field now has alternative name "GLEKJMEEOMH"
    * Note: This field enumerates dungeon entries for different days of week.
* Implement dungeon replay flow by kicking every player out of its scene and adding them back in again.
    * This causes the scene to be cleaned up so that scripting states are restarted when the players are spawned back in. Otherwise the Key trigger to start the dungeon challenge will be missing, as will the barrier to the reward statue.
* Fix GadgetRewardStatue not actually dispensing rewards.
    * Seems the dungeons used to have DungeonChallenge, but they are now WorldChallenge. GadgetRewardStatue was only dispensing rewards for DungeonChallenge.
        * DungeonChallenge is currently used only in Tower.
* Implement UnfreezeGroupLimit.
    * This fixes unlocking the first weapon mats dungeon and hopefully other dungeons with similar unlock requirements.
    * These dungeons have attribute `groupLimit: true` in `BinOutput/Scene/Point/scene*_point.json`. Lua scripts would call UnfreezeGroupLimit, which should then send an UnfreezeGroupLimitNotify to trigger the unlock.

### Under the hood

* Update ScriptLib to pass in current group ID in addition to config ID. Getting the first entity found with only a config ID can lead to wrong outcomes. Config IDs are not unique in a scene.

## Issues fixed by this PR

https://github.com/Grasscutters/Grasscutter/issues/2282

## Type of changes

- [x] Bug fix
- [x] New feature
- [x] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/command/commands/DebugCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/DebugCommand.java
@@ -32,9 +32,12 @@ public final class DebugCommand implements CommandHandler {
 
                 var scene = sender.getScene();
                 var entityId = Integer.parseInt(args.get(0));
+                // TODO Might want to allow groupId specification,
+                // because there can be more than one entity with
+                // the given config ID.
                 var entity =
                         args.size() > 1 && args.get(1).equals("config")
-                                ? scene.getEntityByConfigId(entityId)
+                                ? scene.getFirstEntityByConfigId(entityId)
                                 : scene.getEntityById(entityId);
                 if (entity == null) {
                     sender.dropMessage("Entity not found.");
--- a/src/main/java/emu/grasscutter/command/commands/EntityCommand.java
+++ b/src/main/java/emu/grasscutter/command/commands/EntityCommand.java
@@ -51,7 +51,10 @@ public final class EntityCommand implements CommandHandler {
         }
 
         param.scene = targetPlayer.getScene();
-        var entity = param.scene.getEntityByConfigId(param.configId);
+        // TODO Might want to allow groupId specification,
+        // because there can be more than one entity with
+        // the given config ID.
+        var entity = param.scene.getFirstEntityByConfigId(param.configId);
 
         if (entity == null) {
             CommandHandler.sendMessage(sender, translate(sender, "commands.entity.not_found_error"));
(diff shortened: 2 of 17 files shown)


YetAG:
I think it is a good PR.
However, I was indeed able to access Domain and receive the reward, but I can't choose between using the concentrated resin or using the resin as is.
I use the resin directly as it is.

KingRainbow44 (approved):

longfruit:
> I think it is a good PR. However, I was indeed able to access Domain and receive the reward, but I can't choose between using the concentrated resin or using the resin as is. I use the resin directly as it is.

Thanks for testing the changes. 

Currently, on interaction with the tree, the client always sends a GadgetInteractReq with resin explicitly set to Original Resin.

It seems something, like an open state or the tree's EntityGadgetInfo field, must be set so that the prompt to select between Original and Condensed appears. I've had no luck finding what would work so far.

PR #2413, opened 2023-10-28
Add events to support scene group substitution

longfruit:
## Description

Add new events to support scene group substitution:

| Event | Called when |
|-|-|
| SceneBlockLoadedEvent | SceneBlock finishes reading from block script |
| SceneMetaLoadEvent | SceneScriptManager starts loading SceneMeta |

Example plugin using these events: [SceneHook](https://github.com/longfruit/SceneHook). Disclaimer: I wrote it.

gif

Basic flow:
1. Plugin provides replacement groups.
2. Replacement groups can be registered on demand or automatically at plugin load time.
    1. On SceneBlock load, plugin can change existing and add new groups.
    2. On SceneMeta load, plugin can hook in dynamic group loads when scene finishes loading.
4. Groups become replaced on demand or automatically at scene load time.

---

Also, fix a script issue caused by EVENT_ANY_MONSTER_DIE firing at the same time.

| Before | After |
|-|-|
|gif|gif|

## Type of changes

- [x] Bug fix
- [x] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/server/event/game/SceneBlockLoadedEvent.java
+++ b/src/main/java/emu/grasscutter/server/event/game/SceneBlockLoadedEvent.java
@@ -1,0 +1,14 @@
+package emu.grasscutter.server.event.game;
+
+import emu.grasscutter.server.event.types.ServerEvent;
+import emu.grasscutter.scripts.data.SceneBlock;
+
+public final class SceneBlockLoadedEvent extends ServerEvent {
+    public SceneBlock block;
+
+    public SceneBlockLoadedEvent(SceneBlock block) {
+        super(Type.GAME);
+
+        this.block = block;
+    }
+}
--- a/src/main/java/emu/grasscutter/server/event/game/SceneBlockLoadedEvent.java
+++ b/src/main/java/emu/grasscutter/server/event/game/SceneBlockLoadedEvent.java
@@ -2,9 +2,11 @@ package emu.grasscutter.server.event.game;
 
 import emu.grasscutter.server.event.types.ServerEvent;
 import emu.grasscutter.scripts.data.SceneBlock;
+import lombok.*;
 
+@Getter
 public final class SceneBlockLoadedEvent extends ServerEvent {
-    public SceneBlock block;
+    private SceneBlock block;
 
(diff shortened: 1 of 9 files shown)


src/main/java/emu/grasscutter/server/event/game/SceneBlockLoadedEvent.java
```diff
@@ -0,0 +1,14 @@
+package emu.grasscutter.server.event.game;
+
+import emu.grasscutter.server.event.types.ServerEvent;
+import emu.grasscutter.scripts.data.SceneBlock;
+
+public final class SceneBlockLoadedEvent extends ServerEvent {
+    public SceneBlock block;
```
KingRainbow44: consider making final or using `@Getter` and `@Setter`

src/main/java/emu/grasscutter/server/event/game/SceneMetaLoadEvent.java
```diff
@@ -0,0 +1,15 @@
+package emu.grasscutter.server.event.game;
+
+import emu.grasscutter.game.world.Scene;
+import emu.grasscutter.server.event.types.ServerEvent;
+
+public final class SceneMetaLoadEvent extends ServerEvent {
+    public Scene scene;
+    public boolean hasOverride;
```
KingRainbow44: consider making final or using `@Getter` and `@Setter`

KingRainbow44 (changes requested):

src/main/java/emu/grasscutter/server/event/game/SceneBlockLoadedEvent.java
longfruit: thanks; changed members to private and used `@Getter`/`@Setter`

src/main/java/emu/grasscutter/server/event/game/SceneMetaLoadEvent.java
longfruit: thanks; changed members to private and used `@Getter`/`@Setter`

KingRainbow44 (approved):

PR #2424, opened 2023-11-04
Only deduct energy when elemental burst actually fires

### LINKED ISSUE #1248: [Bug] Energy will be cleared but elemental burst will not be released

**Describe the bug**

If we press the Q key after the E key quickly, energy will be cleared but the burst will not be released. Almost all avatars encountered this issue. 

**Which branch did you use?**

**Screenshots**

1906bf2c5d12aec04e70538c8e35e763_1

**Additional context**

longfruit:
## Description

Elemental burst (Q) can misfire occasionally if done when dashing, doing elemental skill (E), or switching avatars. The client would still send EvtDoSkillSuccNotify, but the burst may not actually happen, and GC currently clears energy whether the burst happened or not.

Fix this by checking AbilityInvocationsNotify packets from the client. On successful bursts, there will be packets with information specific to the burst, like setting resistance/invincible state, AvatarSkillStart action. These can be used to decide when to clear energy and set invulnerability.

Also fix Wanderer's energy not clearing if bursting while floating.

---

This fix could use more testing. I can't exhaustively test through all characters and scenarios. But at least my quick swap teams in abyss no longer have any burst issue.

Preview:
* Faruzan: Works fine
* Bennett: E into Q causes misfire
* Yelan: E into Q causes misfire
* Wanderer: E into Q doesn't clear energy

| Before | After |
|-|-|
| gif | gif |

## Issues fixed by this PR

https://github.com/Grasscutters/Grasscutter/issues/1248

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/data/binout/AbilityModifier.java
+++ b/src/main/java/emu/grasscutter/data/binout/AbilityModifier.java
@@ -327,6 +327,7 @@ public class AbilityModifier implements Serializable {
         public String srcKey, dstKey;
 
         public int skillID;
+        public int resistanceListID;
 
         public AbilityModifierAction[] actions;
         public AbilityModifierAction[] successActions;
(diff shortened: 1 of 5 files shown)


KingRainbow44:
ITS THE BUG FIX

KingRainbow44 (approved):
bug fix for a bug from a year ago pog

PR #2449, opened 2023-12-02
Fix some more dungeons

longfruit:
* Monds weapon mats domain: Fix time between kill not refreshing
* Inaz husk domain: Fix broken domain challenge
    * `EVENT_ANY_MONSTER_LIVE` is likely sent on tick, not on create. See scene40801_group240801001.lua:
         1. `condition_EVENT_ANY_MONSTER_LIVE_1023` checks for mob 1008 to spawn AND for variable `challenge` to be 1 
         2. Mob 1008 spawns during `action_EVENT_SELECT_OPTION_1003`, at `ScriptLib.AddExtraGroupSuite(context, 240801001, 2)`
         3. This spawn triggers `EVENT_ANY_MONSTER_LIVE` for mob 1008 but still fails the condition because `challenge` is still 0.
         4. `challenge` is set to 1 at the end of `action_EVENT_SELECT_OPTION_1003`. By now, `EVENT_ANY_MONSTER_LIVE` for mob 1008 no longer fires, causing the domain challenge to fail to start.

| gif | gif |
|-|-|

## Description

## Type of changes

- [x] Bug fix
- [ ] New feature 
- [ ] Enhancement
- [ ] Documentation

## Checklist:

### DIFF
--- a/src/main/java/emu/grasscutter/game/dungeons/challenge/WorldChallenge.java
+++ b/src/main/java/emu/grasscutter/game/dungeons/challenge/WorldChallenge.java
@@ -29,7 +29,7 @@ public class WorldChallenge {
     private final AtomicInteger score;
     private boolean progress;
     private boolean success;
-    private long startedAt;
+    private int startedAt;
     private int finishedTime;
 
     /**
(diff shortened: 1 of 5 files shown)


Hartie95:
It's possible that this is correct, but it's also possible that instead the events from the spawn should be handled only after the initial lua execution finished, in which case it would be more something like a race condition instead of a wrong way to trigger the event. 
We should probaly keep an eye on the live events.

longfruit:
> the events from the spawn should be handled only after the initial lua execution finished

The thought crossed my mind as well. Perhaps each suite group should have only one lua execution in flight at a time, so that lua events from gadget state change or mob creation by the same trigger would queue up and fire afterwards.

I'll experiment more if onTick live events become a perf issue or we see the same race condition with gadget state changes or something similar to mob live event.

KingRainbow44 (approved):

PR #2455, opened 2023-12-14
Added docker support

### LINKED ISSUE #2018: Complete construction steps（求linux完整搭建过程，centos版本）
Complete construction steps（求linux完整搭建过程，centos版本）
If yes, please send it to the mailbox（如果有请发送到邮箱）
[1773232143@qq.com](mailto:1773232143@qq.com)
Thank you.

Maybe we can discuss it here

### LINKED ISSUE #2038: [Feature Request] 可不可以添加Docker一键部署？
将所有环境，包括mongo打包成一个docker，这样的话，仅安装一个docker就可以把服务端跑起来了，docker中包含mongo，jdk1.7等

### LINKED ISSUE #2250: [Feature Request] 大大们有没有考虑过出docker版
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

部署更方便点，也可以兼容更多环境

**Describe alternatives you've considered**

**Additional context**

YannickFricke:
## Description

- Added the multi-staged Dockerfile
- Added the docker-compose.yml file
  - It also includes a MongoDB service
- Added the entrypoint for the Docker image

## Issues fixed by this PR

Closes #2038 
Closes #2250
Closes #2018 
## Type of changes

- [ ] Bug fix
- [ ] New feature 
- [X] Enhancement
- [ ] Documentation

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] My pull request is unique and no other pull requests have been opened for these changes


KingRainbow44 (changes requested):
while I think generating the config from a JS script is good, it would be a bit of a hassle to maintain. is there a specific reason you can't generate a config using Java and reflection? additionally, it would be nice to have an environment variable to change where the image gets its resources.

YannickFricke:
> it would be a bit of a hassle to maintain

I think the same about the configuration file - it would be far better if GrassCutter would use environment variables directly - so we could get rid of the workaround (even without having all the reflection stuff) :)

My current approach was:
1. Build the server from scratch
2. Running the container via docker-compose (luckily the MongoDB adapter has a timeout of 30 seconds)
3. `cat` the config to the host system

Sadly GrassCutter doesn't provide a default configuration file in the source code :/

> is there a specific reason you can't generate a config using Java and reflection

For me particular I've never worked with gradle - so we would need another entrypoint where also GSON is loaded - so we could easily generate the config.

But as said above - it would be far better if GrassCutter would use environment variables.

> additionally, it would be nice to have an environment variable to change where the image gets its resources.

You mean the GIT repository? That would be doable :)

YannickFricke:
It also seems that there is another problem with the configuration file:

grafik

In the code it states that it is at version 13 - but the server only generates a config of version 4...

YannickFricke:
Also if we would refactor GC to use environment variables - this would be a breaking change :)

YannickFricke:
@KingRainbow44 Could you please verify the latest changes if they work for you? :)

I've refactored the ConfigContainer to use environment variables instead of a file configuration :)

The last TODO is only to refactor the Dockerfile to use an "ARG" for the resources URL - but thats a no-brainer :)

YannickFricke:
The server is starting correctly in Docker (even with the MongoDB connection) - but I cant test it with the client (need to download it first with this slow connection here)

grafik

KingRainbow44:
how does this read from the traditional 'config.json' file?

YannickFricke:
> how does this read from the traditional 'config.json' file?

It does not anymore - everything was converted to environment variables.

Due to the lack of Docker supporting only mounting specific files (AFAIK) you cannot even mount the config into the container under a specific path.

KingRainbow44:
in it's current state, i don't think this pr will be going through. dropping _complete_ support for `config.json` would be a disaster for the many users which already use grasscutter.

YannickFricke:
> in it's current state, i don't think this pr will be going through. dropping _complete_ support for `config.json` would be a disaster for the many users which already use grasscutter.

What about reading out the config.json file (like before) and generating a script (.bat / .sh) based on the old configuration and then deleting the file (so the generated script wont be overwritten)?

KingRainbow44:
not everyone uses a script to start grasscutter, instead why can't you read from a `config.json` if it exists and set environment variables like that?

YannickFricke:
> not everyone uses a script to start grasscutter, instead why can't you read from a `config.json` if it exists and set environment variables like that?

Because it would bloat up the Docker image - you would need dependencies like jq and an extensive bash script which are only used once in the whole lifecycle.

And we all know that data transformations (which would be needed for nested config keys) isn't that comfortable to maintain.

PR #2486, opened 2024-03-17
Add Docker Support

kashalls:
## Description
This PR extends off of #2455, simply adding support for building the required containers and publishing them to ghcr.io

This PR introduces a github workflow to build and publish a docker container that runs grasscutter on `amd64` platforms. The base builder container that gradle provides isn't supported for arm64, therefore we can't build for Raspberry Pi's/ARM-based platforms. This might be fixed in the future.

Configuration file can be loaded by using `-v C:\\Grasscutter\\test.json:/app/config.json`

## Issues fixed by this PR
See #2455 

## Type of changes
- [ ] Bug fix
- [ ] New feature 
- [X] Enhancement
- [ ] Documentation

## Checklist:

- [ ] My code follows the style guidelines of this project
- [X ] My pull request is unique and no other pull requests have been opened for these changes

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -46,6 +46,6 @@ jobs:
         with:
           context: .
           push: true
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,0 +1,51 @@
+---
+name: Build Docker Container
+on:
+  push:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout Project
+        uses: actions/checkout@v4
+
+      - name: Generate Docker Meta
+        uses: docker/metadata-action@v5
+        id: meta
+        with:
+          images: ghcr.io/kashalls/grasscutter
+          tags: |
+            type=ref,event=branch
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=sha
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+RUN git clone --branch ${DATA_BRANCH} --depth 1 ${DATA_REPOSITORY}
+
+# Result Container
+FROM amazoncorretto:17-alpine
+
+WORKDIR /app
+
+# Copy built assets
+COPY --from=builder /app/grasscutter-1.7.4.jar /app/grasscutter.jar
```
KingRainbow44: this doesn't seem good; hardcoding the version of the jar (which is bound to change) will break this in the future

.github/workflows/build-container.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    steps:
+      - name: Checkout Project
+        uses: actions/checkout@v4
+
+      - name: Generate Docker Meta
+        uses: docker/metadata-action@v5
+        id: meta
+        with:
+          images: ghcr.io/kashalls/grasscutter
```
KingRainbow44: reference to your own user?

KingRainbow44 (changes requested):

Dockerfile
kashalls: Fixed by replacing the version with a wildcard. Assumes that their will only be one file produced that matches `grasscutter-*.jar`.

.github/workflows/build-container.yml
kashalls: Replaced with `${{ github.repository }}` which returns [org_name/org_repo](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

Hartie95:
Just out of curiosity, why is the npm/nodejs update /install required in the docker script?

KingRainbow44:
presumably for handbook generation; but since it requires extra data, it was skipped anyways

kashalls:
> Just out of curiosity, why is the npm/nodejs update /install required in the docker script?

I added it for handbook generation, makes it easy to add the data into the data task and include it if its needed.

gptlang:
Why no docker-compose?

kashalls:
> Why no docker-compose?

Docker compose still requires a docker image. If you want to add it, go for it.

PR #2524, opened 2024-08-02
Fixes for multi version support

Hartie95:
This replaces all hardcoded opcode usages outside of proto packets and handlers using the old system


Hartie95:
ops

KingRainbow44:
this looked like a fun PR to review

RealistikDash:
GitHub defaulting to opening a pr on the original project is the worst curse known to man.

